PR #130, opened 2019-07-26
move static icu dependencies to FindICU so it can be overruled from outside

dothebart:


gnusi:
fixed in different PR

PR #150, opened 2019-12-23
Bug fix/removing dangling cached readers

Dronplane:
During consolidation readers cache is populated. If consolidation is finished during commit (additional removes are applied) newly created segment will immediately change version (while old version is in readers cache). This old version should be added to segments mask for purging from cache despite  of this segment is not empty (it is already outdated!)

### DIFF
--- a/core/index/index_writer.cpp
+++ b/core/index/index_writer.cpp
@@ -2085,6 +2085,11 @@ index_writer::pending_context_t index_writer::flush_all(const before_commit_f& b
 
     // write non-empty document mask
     if (!docs_mask.empty()) {
+      if (!pending_consolidation) { // if this is pending consolidation this segment is already in the mask (see assert below)
+        //new version will be created created. Remove old version from cache!
+        ctx->segment_mask_.emplace(pending_segment.segment.meta);
+      }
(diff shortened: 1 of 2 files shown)


core/index/index_writer.cpp
```diff
@@ -2085,6 +2085,11 @@ index_writer::pending_context_t index_writer::flush_all(const before_commit_f& b
 
     // write non-empty document mask
     if (!docs_mask.empty()) {
+      if (!pending_consolidation) { // if this is pending consolidation this segment is already in the mask (see assert below)
+        //new version will be created. Remove old version from cache!
```
gnusi:
```suggestion
        // new version will be created. Remove old version from cache!
```

core/index/index_writer.cpp
```diff
@@ -2085,6 +2085,11 @@ index_writer::pending_context_t index_writer::flush_all(const before_commit_f& b
 
     // write non-empty document mask
     if (!docs_mask.empty()) {
+      if (!pending_consolidation) { // if this is pending consolidation this segment is already in the mask (see assert below)
```
gnusi:
```suggestion
      if (!pending_consolidation) { 
        // if this is pending consolidation, 
        // this segment is already in the mask (see assert below)
```

PR #166, opened 2020-03-16
fix issue with optimizing of wildcard filters

gnusi:


gnusi:
https://jenkins.arangodb.biz/view/iResearch/job/iresearch-release-s/994/

PR #202, opened 2020-08-19
Add payload to segment created after index_writer::clear

Dronplane:
Moves segment meta payload functor to index_writer params.
Adds payload for segment created after index is cleared

Fix is verified by gtest

http://jenkins.arangodb.biz:8080/view/PR/job/iresearch_s-pr/136/


Dronplane:
Tests blue except known locale failures on windows

PR #203, opened 2020-08-24
bug-fix/internal-issue-#747 

Dronplane:
PR fixes 2 cases:
1. When newly inserted document is removed before commit has passed - dangling cached reader is left pointing to old segment version. This version actually will be never committed to index meta (new version, with removes will be committed). Now if newly created segment has removes - it is masked from cache.

2. When all newly created segments are cleared by removes, pending context is dropped leaving readers in cache. Now if at flush_all exit segment mask is not empty - it is applied for cache purging on exit.

Fix is verified by gtests

http://jenkins.arangodb.biz:8080/view/PR/job/iresearch_s-pr/137/

### DIFF
--- a/core/index/index_writer.cpp
+++ b/core/index/index_writer.cpp
@@ -1411,7 +1411,7 @@ bool index_writer::consolidate(
 
   std::set<const segment_meta*> candidates;
   const auto run_id = reinterpret_cast<size_t>(&candidates);
-  
+
   // hold a reference to the last committed state to prevent files from being
   // deleted by a cleaner during the upcoming consolidation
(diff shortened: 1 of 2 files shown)


Dronplane:
Debug run to verify assertions 

http://jenkins.arangodb.biz:8080/job/iresearch-debug/1019/

Dronplane:
Correct static build
http://jenkins.arangodb.biz:8080/view/iResearch/job/iresearch-debug-s/922/

Dronplane:
http://jenkins.arangodb.biz:8080/view/PR/job/iresearch_s-pr/139/

gnusi (approved):

PR #204, opened 2020-09-17
Feature/pipeline analyzer

Dronplane:
PR add new type of analyzer - pipeline analyzer. This is basically a container for other analyzers. This pipeline is applied sequentially to input data.

https://jenkins.arangodb.biz/view/PR/job/iresearch_s-pr/140/

### DIFF
--- a/core/analysis/pipeline_token_stream.cpp
+++ b/core/analysis/pipeline_token_stream.cpp
@@ -186,8 +186,10 @@ pipeline_token_stream::pipeline_token_stream(const pipeline_token_stream::option
 	bottom_ = --pipeline_.end();
 }
 
-
-/// Pipeline position change rules:
+/// Moves pipeline to next token.
+/// Term is taken from last analyzer in pipeline
+/// Offset is recalculated accordingly
+/// Increment is calculated according to following position change rules
 ///  - If none of pipeline members changes position - whole pipeline holds position
 ///  - If one or more pipeline member moves - pipeline moves( change from max->0 is not move, see rules below!).
 ///    All position gaps are accumulated (e.g. if one member has inc 2(1 pos gap) and other has inc 3(2 pos gap)  - pipeline has inc 4 (1+2 pos gap))
@@ -198,14 +200,14 @@ pipeline_token_stream::pipeline_token_stream(const pipeline_token_stream::option
 ///    position from any positive value back to 0 due to reset (additional gaps also preserved!) as this is
 ///    not change max->0 and position is indeed changed.
 inline bool pipeline_token_stream::next() {
-	uint32_t upstream_inc = 0;
+	
 	while (!current_->next()) {
 		if (current_ == top_) { // reached pipeline top and next has failed - we are done
 			return false;
 		}
 		--current_;
 	}
-	upstream_inc += current_->inc->value;
+	uint32_t pipeline_inc{ current_->inc->value };
 
 	const auto top_holds_position = current_->inc->value == 0;
 
@@ -226,15 +228,15 @@ inline bool pipeline_token_stream::next() {
 			}
 			--current_;
 		}
-		upstream_inc += current_->inc->value;
+		pipeline_inc += current_->inc->value;
 		assert(current_->inc->value > 0); // first increment after reset should be positive to give 0 or next pos!
-		assert(upstream_inc > 0);
-		upstream_inc--; // compensate placing sub_analyzer from max to 0 due to reset
+		assert(pipeline_inc > 0);
+		pipeline_inc--; // compensate placing sub_analyzer from max to 0 due to reset
 										// as this step actually does not move whole pipeline
-										// sub analyzer just stays same pos as it`s parent (rollback step will be done below if necessary!)
+										// sub analyzer just stays same pos as it`s parent (step for rollback to 0 will be done below if necessary!)
 	}
 	if (step_for_rollback) {
-		upstream_inc++;
+		pipeline_inc++;
 	}
 
 	// FIXME: get rid of full recalc. Use incremental approach
@@ -242,13 +244,13 @@ inline bool pipeline_token_stream::next() {
 	uint32_t upstream_end{ static_cast<uint32_t>(pipeline_.front().data_size) };
 	for (const auto& p : pipeline_) {
 		start += p.offs->start;
-		if (p.offs->end != p.data_size && p.analyzer != bottom_->analyzer) {
+		if (p.offs->end != p.data_size && p.analyzer != bottom_->analyzer) { // TODO: hide analyzer and remove kludge check!
 			// this analyzer is not last and eaten not all its data.
 			// so it will mark new pipeline offset end.
 			upstream_end = start +  (p.offs->end - p.offs->start);
 		}
 	}
-	inc_.value = upstream_inc;
+	inc_.value = pipeline_inc;
 	offs_.start = start;
 	offs_.end = current_->offs->end == current_->data_size ? 
 		          upstream_end : // all data eaten - actual end is defined by upstream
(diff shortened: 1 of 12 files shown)


core/analysis/pipeline_token_stream.cpp
```diff
@@ -0,0 +1,295 @@
...
+  if (step_for_rollback) {
+    pipeline_inc++;
+  }
+  inc_.value = pipeline_inc;
+  offs_.start = current_->start();
+  offs_.end = current_->end();
+  return true;
+}
+
+inline bool pipeline_token_stream::reset(const string_ref& data) {
```
gnusi, line 284: why `inline` in `cpp`?

core/analysis/pipeline_token_stream.cpp
```diff
@@ -0,0 +1,295 @@
...
+///  - If none of pipeline members changes position - whole pipeline holds position
+///  - If one or more pipeline member moves - pipeline moves( change from max->0 is not move, see rules below!).
+///    All position gaps are accumulated (e.g. if one member has inc 2(1 pos gap) and other has inc 3(2 pos gap)  - pipeline has inc 4 (1+2 pos gap))
+///  - All position changes caused by parent analyzer move next (e.g. transfer from max->0 by first next after reset) are collapsed.
+///    e.g if parent moves after next, all its children are resetted to new token and also moves step froward - this whole operation
+///    is just one step for pipeline (if any of children has moved more than 1 step - gaps are preserved!)
+///  - If parent after next is NOT moved (inc == 0) than pipeline makes one step forward if at least one child changes
+///    position from any positive value back to 0 due to reset (additional gaps also preserved!) as this is
+///    not change max->0 and position is indeed changed.
+inline bool pipeline_token_stream::next() {
```
gnusi, line 235: remove `inline`

core/analysis/pipeline_token_stream.cpp
```diff
@@ -0,0 +1,295 @@
...
+  normalize_json_config);
+
+NS_END
+
+NS_ROOT
+NS_BEGIN(analysis)
+
+pipeline_token_stream::pipeline_token_stream(const pipeline_token_stream::options_t& options)
+  : attributes{ {
+    { irs::type<increment>::id(), &inc_ },
```
gnusi, line 209: also there is optional `payload` attribute

core/analysis/pipeline_token_stream.hpp
```diff
@@ -0,0 +1,107 @@
...
+      }
+      return false;
+    }
+
+    uint32_t start() const noexcept {
+      return data_start + offs->start;
+    }
+
+    uint32_t end() const noexcept {
+      return offs->end == data_size ?
```
gnusi, line 78: offset can be `nullptr`

core/analysis/pipeline_token_stream.hpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+    const term_attribute* term;
+    const increment* inc;
+    const offset* offs;
+    size_t data_size{ 0 };
+    uint32_t data_start{ 0 };
+    uint32_t data_end{ 0 };
+    uint32_t pos{ irs::integer_traits<uint32_t>::const_max };
+
+   private:
```
gnusi, line 90: having private members in a private struct looks a bit weird?

core/analysis/pipeline_token_stream.hpp
```diff
@@ -0,0 +1,107 @@
...
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is ArangoDB GmbH, Cologne, Germany
+///
+/// @author Andrei Lobov
+////////////////////////////////////////////////////////////////////////////////
+
+#ifndef IRESEARCH_IQL_PIPELINE_TOKEN_STREAM_H
```
gnusi, line 24: rename to `IRESEARCH_PIPELINE_TOKEN_STREAM_H`

core/analysis/pipeline_token_stream.hpp
```diff
@@ -0,0 +1,107 @@
...
+  virtual bool next() override;
+  virtual bool reset(const string_ref& data) override;
+
+ private:
+  struct sub_analyzer_t {
+    explicit sub_analyzer_t(const irs::analysis::analyzer::ptr& a)
+      : term(irs::get<irs::term_attribute>(*a)),
+        inc(irs::get<irs::increment>(*a)),
+        offs(irs::get<irs::offset>(*a)),
+        analyzer(a) {}
```
gnusi, line 56: `assert` that `increment` isn't a null pointer

core/analysis/pipeline_token_stream.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+REGISTER_ANALYZER_JSON(irs::analysis::pipeline_token_stream, make_json,
+  normalize_json_config);
+
+NS_END
+
+NS_ROOT
+NS_BEGIN(analysis)
+
+pipeline_token_stream::pipeline_token_stream(const pipeline_token_stream::options_t& options)
```
gnusi, line 207: `pipeline_token_stream` must accept `std::vector<irs::analysis::analyzer::ptr>&&` to ensure that pipeline analyzers **owns** sub-analyzers, otherwise a one analyzer could be used in different pipelines which is error prone.

core/analysis/pipeline_token_stream.cpp
```diff
@@ -0,0 +1,295 @@
...
+  : attributes{ {
+    { irs::type<increment>::id(), &inc_ },
+    { irs::type<offset>::id(), &offs_ },
+    { irs::type<term_attribute>::id(), irs::get_mutable<term_attribute>(options.pipeline.back().get()) }},
+    irs::type<pipeline_token_stream>::get() } {
+  pipeline_.reserve(options.pipeline.size());
```
gnusi, line 213: let's add an assertion to ensure that input analyzers don't contain null pointers

core/analysis/pipeline_token_stream.cpp
```diff
@@ -0,0 +1,295 @@
...
+  for (const auto& p : options.pipeline) {
+    pipeline_.emplace_back(p);
+  }
+  assert(!pipeline.empty());
```
gnusi, line 217: in case of empty input array let's set `top_` and `bottom_` to a some invalid analyzer which does nothing but returns `false` on every `next()`

core/analysis/pipeline_token_stream.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+NS_END
+
+NS_ROOT
+NS_BEGIN(analysis)
+
+pipeline_token_stream::pipeline_token_stream(const pipeline_token_stream::options_t& options)
+  : attributes{ {
+    { irs::type<increment>::id(), &inc_ },
+    { irs::type<offset>::id(), &offs_ },
```
gnusi, line 210: we have to think about how to handle analyzers exposing different attributes. Our inversion procedure assumes that `term_attribute` and `increment` are mandatory attributes, while `offset` and `payload` are optional.
It's also a bit unclear how to handle `payload` itself.

gnusi (changes requested):

core/analysis/pipeline_token_stream.cpp
Dronplane: I suggest the following:
 - if any pipeline member has no `offset` attribute than whole pipeline has no `offset` as it is impossible to calculate final offset in general. IF all members has `offset` than it is calculated as implemented.
- `Payload` is taken from lowes pipeline member. As we clearly could not provide several payloads per token -we have to choose. I suggest the closest to "token".

core/analysis/pipeline_token_stream.hpp
Dronplane: It is to protect from accidental direct calls to sub-analyzer next/reset. This will break proper offset/position tracking. Test will surely catch it, but prefer to detect error on compile time.

Dronplane:
https://jenkins.arangodb.biz/job/iresearch_s-pr/141/

gnusi (approved):
LGTM

PR #206, opened 2020-10-05
Feature/immutable-fst

gnusi:


gnusi:
http://jenkins01.arangodb.biz:8080/view/PR/job/iresearch_s-pr/149/

gnusi:
http://jenkins01.arangodb.biz:8080/view/PR/job/iresearch_s-pr/150/

core/formats/formats_burst_trie.cpp
```diff
@@ -54,30 +58,286 @@
...
+  }
+
+  void clear() {
+    str_.clear();
+    ref_ = ref_t::NIL;
+  }
+
+  template<bool Volatile>
+  void assign(const ref_t& str) {
+    if (Volatile) {
```
Dronplane, line 127:
```suggestion
    if constexpr (Volatile) {
```

core/formats/formats_burst_trie.cpp
```diff
@@ -54,30 +58,286 @@
...
+      str_.assign(str.c_str(), str.size());
+      ref_ = str_;
+    } else {
+      ref_ = str;
```
Dronplane, line 131: to reduce memory consumption in case of switches between volatile/non-volatile
```suggestion
      ref_ = str;
      str_.clear();
```

core/formats/formats_burst_trie.cpp
```diff
@@ -54,30 +58,286 @@
...
+  volatile_ref(volatile_ref&& rhs) noexcept
+   : str_(std::move(rhs.str_)),
+     ref_(str_.empty() ? rhs.ref_ : ref_t(str_)) {
+    rhs.ref_ = ref_;
+  }
+
+  volatile_ref& operator=(volatile_ref&& rhs) noexcept {
+    if (this != &rhs) {
+      str_ = std::move(rhs.str_);
+      ref_ = (str_.empty() ? rhs.ref_ : ref_t(str_));
```
Dronplane, line 114: Is it safe to assume that if str_ is empty it is ref to external ? Especially if non-volatile assign works as implemented (not cleaning str_). It would be more safe to check initially if rhs.ref_ points to rhs.str_.

Dronplane (approved):
LGTM

PR #214, opened 2020-11-12
Be able to find and use ICU source dir

elfringham:
The source file ucnv.c was renamed to ucnv.cpp in 2016, so update to use it. Also newer cmake needs a policy set to use xxx_ROOT variables to locate source directories.


gnusi:
Hi @elfringham,
thanks for the contribution. As this project is being maintained by arangodb, I have to ask for a CLA (https://www.arangodb.com/documents/cla.pdf) in order to get this PR merged. More information here.

elfringham:
@gnusi I have returned the signed CLA.

joerg84:
> @gnusi I have returned the signed CLA.

Thanks @elfringham . Confirming the receipt!

elfringham:
Updated with a 2nd commit to ensure it still builds on cmake versions without the new policy.

gnusi:
Hey @elfringham,
just curious what are doing with `iresearch`?

elfringham:
I work for Linaro and have been asked to see if I can get ArangoDB to build for AARCH64. My first issue has been the use of x86 intrinsics in IResearch so am working to get USE_SIMDCOMP to be honoured correctly in the build of IResearch, which I hope will then get merged into the ArangoDB build in due course. Longer term I would like to have ARM SVE code as an alternative in places where x86 intrinsics are used so that AARCH64 users can also enjoy improved performance.
But my first steps to this end is just to see if I can sort out some of the the reasons why the CI checks fail. I would hate to make any changes that introduced any sort of regression, so think that working CI checks are very valuable to all.
There seems to have been the ability to build ArangoDB for ARM previously but it seems that this is not currently the case.

gnusi:
@elfringham Good to know. I suggest to start fixing issues directly in `arango` repo as later we can backport necessary fixes to `iresearch`.

Off the top of my head there are at least two potential issues on `iresearch` side:
* SIMD intrinsics
SIMD can be easily turned off by switching to different storage format, so it should not be a problem for runtime
* `cmpxchg16`

Anyway, I suggest you join our slack as we can discuss everything in more convenient fashion :-)
https://join.slack.com/t/arangodb-community/shared_invite/zt-d3a8xnj8-Uf66flfhU8pQARwjOXx~Tg

gnusi:
@elfringham That was already addressed in ef170a8dea6922bb3ef1556ed4d3ea28442d1a6f

PR #218, opened 2020-11-14
fix compile warning with clang11

jsteemann:
```
In file included from ./arangod/IResearch/GeoFilter.cpp:34:
./3rdParty/iresearch/core/search/disjunction.hpp:1295:24: warning: lambda capture 'this' is not used [-Wunused-lambda-capture]
      visit_and_purge([this, &empty](auto& it) mutable {
```


gnusi:
@jsteemann 
I don't think that's right, as lamda calls "refill" member

PR #219, opened 2020-11-14
Fix for build with libunwind

elfringham:
Fix link errors when libunwind is available due to incorrect function signatures.


elfringham:
The checks for this look worse than they are in reality as the pull request was created before the yiddish fix was merged.

PR #220, opened 2020-11-14
Fix build of tests with shared objects

elfringham:
Fix build of tests with shared objects


elfringham:
Updated with a 2nd commit to add some build dirs to library search path which works in my repo, but as this pull request was created before the yiddish fix was merged, the checks look worse than they are in reality.

Dronplane:
Sorry we could not accept this PR. We have some issues with shared builds in general (especially for windows). So there are plans to  rewrite this part heavily anyway.

PR #224, opened 2020-11-17
Fix build for gcc-7 and gcc-8

elfringham:


gnusi:
@elfringham according to the standard `std::set(std::set&&)` is allowed to throw thus any type containing `std::set<...>` as a member can't have `noexcept` move constructor/assignment (unless you don't need to move `std::set<...>` member ofc). Marking move constructor/assignment only makes things worse as it overrides deduced `noexcept` value. That may cause process termination if `std::set<...>` move constructor/assignment actually throws an exception.
I'd appreciate if we can chat to understand your use case/interest better.

elfringham:
That's fine. I just see the assert failure in the builds for gcc-7 and gcc-8 during the CI checks. Maybe those builds should just be disabled or the assert removed if it can't or even shouldn't be fixed.
Constantly failing CI checks just waste time on the CI servers.

gnusi:
Ok, fixing travis is on our todo list. The thing is our main priority is arango and gcc9, clang11 and msvc2019 as the consequence. Testing against different compilers gives more stability but we don't always have enough time to keep everything 100% green.
I'm going to close this for now then.

PR #228, opened 2020-12-04
feature/external-buffers-for-buffered-streams

gnusi:
Previously buffered streams owned memory buffer they use for serving low level IO operations. That's not very convenient for modern IO API. This PR refactors buffered streams to allow setting an external buffer.

### DIFF
--- a/core/store/data_input.cpp
+++ b/core/store/data_input.cpp
@@ -71,21 +71,6 @@ std::streamsize input_buf::showmanyc() {
 // --SECTION--                               buffered_index_input implementation
 // -----------------------------------------------------------------------------
 
-buffered_index_input::buffered_index_input(
-    size_t buf_size/*= DEFAULT_BUFFER_SIZE*/
-) noexcept
-  : start_(0),
-    buf_size_(buf_size) {
-}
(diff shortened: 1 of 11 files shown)


gnusi:
http://jenkins01.arangodb.biz:8080/view/PR/job/iresearch_s-pr/156/

core/store/data_input.cpp
```diff
@@ -172,25 +157,19 @@ size_t buffered_index_input::refill() {
...
 
-  if (!buf_) {
-    buf_ = memory::make_unique<byte_type[]>(buf_size_);
-    begin_ = end_ = buf_.get();
-    seek_internal(start_);
-  }
-
-  begin_ = buf_.get();
-  end_ = begin_ + read_internal(buf_.get(), data_size);
+  begin_ = buf_;
```
Dronplane, line 160:
```suggestion
  assert(buf_);
  begin_ = buf_;
```
As now we have no auto-allocation in case of no buffer so at least we need to assert we have it.

Dronplane (approved):
LGTM. One minor assert suggestion

PR #238, opened 2021-02-02
attempt to reduce suggest-override warnings

jsteemann:
Attempt to fix some warnings that occur when compiling with `-Wsuggest-override`.

Not tested yet, so we should wait for the automated tests to pass.


Dronplane (approved):
LGTM

PR #240, opened 2021-02-22
Feature/speedup-indexing

gnusi:


core/store/memory_directory.cpp
```diff
@@ -538,19 +538,22 @@ bool memory_directory::rename(
...
+        files_.erase(it);
+        return true;
+      }
     }
 
-    files_[dst] = std::move(it->second);
-    files_.erase(it);
-
-    return true;
+    files_.erase(res.first);
```
Dronplane, line 556: if `dst` already present (or `src`  is not found) in `files_` this line will remove `dst` (and leave `src` as is). Is that intentional? Old version was overwriting `dst` in the case of existing `dst` and leaving `dst` as is in case of  `src` was not found .

Dronplane (approved):
LGTM

PR #245, opened 2021-03-13
feature/bitset-iterator-micro-optimization

gnusi:
https://vpna.arangodb.biz:3456/view/PR/job/iresearch_s-pr/200/

### DIFF
--- a/core/search/bitset_doc_iterator.cpp
+++ b/core/search/bitset_doc_iterator.cpp
@@ -22,7 +22,6 @@
 
 #include "bitset_doc_iterator.hpp"
 #include "formats/empty_term_reader.hpp"
-#include "utils/bitset.hpp"
 #include "utils/math_utils.hpp"
 
 namespace iresearch {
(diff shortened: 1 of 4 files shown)


core/search/bitset_doc_iterator.cpp
```diff
@@ -92,8 +97,10 @@ doc_id_t bitset_doc_iterator::seek(doc_id_t target) noexcept {
     break;
   }
 
-  base_ = doc_id_t(std::distance(begin_, next_) * bits_required<word_t>());
-  word_ = (*next_++) & ((~word_t(0)) << bitset::bit(target));
+  const doc_id_t bit_idx = target % bits_required<word_t>();
+  base_ = word_idx * bits_required<word_t>();
+  word_ = (*next_++) >> bit_idx;
+  doc_.value = base_ - 1 + bit_idx;
```
Dronplane, line 103: As a wild idea - maybe just add here piece of code from next? And save some cpu cycles on next call. It`s a code dupication, but we are speeding very low-level part. 
So add something like:
```suggestion
  if(_word) {
    const doc_id_t delta = doc_id_t(math::math_traits<word_t>::ctz(word_));
    word_ = (word_ >> delta) >> 1;
    doc_.value = base_ + bit_idx + delta;
    return doc_value;
  }
```
 But that`s additional conditional branching. So maybe I`m wrong.

Dronplane (approved):
LGTM. Plus one wild idea :)

PR #247, opened 2021-03-19
Feature/async directory

gnusi:


gnusi:
Async directory is still in unstable state, so we leave it turned off by default

Dronplane (approved):
LGTM

PR #249, opened 2021-03-19
Feature/columnstore 2.0

gnusi:


core/store/store_utils.hpp
```diff
@@ -377,40 +397,54 @@ class IRESEARCH_API bytes_ref_input : public index_input {
...
     return pos_ >= data_.end();
   }
 
-  virtual byte_type read_byte() override final {
+  virtual byte_type read_byte() noexcept override final {
     assert(pos_ < data_.end());
     return *pos_++;
   }
 
+  virtual const byte_type* read_buffer(size_t offset, size_t size, BufferHint hint) noexcept override final {
```
Dronplane, line 422:
```suggestion
  virtual const byte_type* read_buffer(size_t offset, size_t size, BufferHint /*hint*/) noexcept override final {
```

core/utils/bytes_utils.hpp
```diff
@@ -79,6 +79,13 @@ struct bytes_io<T, sizeof(uint16_t)> {
     return out;
   }
 
+  static T read(byte_type*& in) {
+    const T out = ((in[0] << 8) | in[1]);
```
Dronplane, line 83: sizeof(T) is expected to be always 2? Otherwise we will read into T 2 bytes but advance by actual size of T which seems odd

Dronplane: Ah indeed its for uint16_t

tests/utils/simd_utils_test.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+#include "utils/misc.hpp"
+#include "utils/std.hpp"
+#include "utils/simd_utils.hpp"
+
+namespace {
+
+}
+
+TEST(simd_utils_test, delta32) {
```
Dronplane, line 33: Should we also test with non-constant difference between values ?

tests/store/store_utils_tests.cpp
```diff
@@ -386,48 +412,39 @@ void vencode_from_array(T expected_value, size_t expected_length) {
...
+void read_write_block_optimized(const std::array<uint32_t, N>& source) {
+  static_assert(128 == N);
+  std::array<uint32_t, N> enc_dec_buf;
+  std::fill_n(enc_dec_buf.data(), N, std::numeric_limits<uint32_t>::max());
+
+  auto pack_block = [](
+      const uint32_t* RESTRICT decoded,
+      uint32_t* RESTRICT encoded,
+      const uint32_t bits) noexcept {
+    std::memset(encoded, 0, sizeof(uint32_t) * N); // FIXME do we need memset???
```
Dronplane, line 425: I vote for removing memset as `simdpackwithoutmask` does not requires output buffer to be zero filled or we should mark it so

core/formats/columnstore2.cpp
```diff
@@ -0,0 +1,1420 @@
...
+    index_input& in) {
+  std::vector<uint64_t> blocks(
+    math::div_ceil32(hdr.docs_count, column::BLOCK_SIZE));
+
+  in.read_bytes(
+    reinterpret_cast<byte_type*>(blocks.data()),
+    sizeof(uint64_t)*blocks.size());
+
+  if constexpr (!is_big_endian()) {
+    // FIXME simd?
```
Dronplane, line 146: 👍

Dronplane (approved):
LGTM

PR #255, opened 2021-04-22
Fix internal issue #798

Dronplane:
fix case when all docs from just consolidated segment are deleted


core/index/index_writer.cpp
```diff
@@ -2159,6 +2159,7 @@ index_writer::pending_context_t index_writer::flush_all() {
     // skip empty segments
     if (!pending_segment.segment.meta.live_docs_count) {
       ctx->segment_mask_.emplace(pending_segment.segment.meta);
+      modified = true;
```
gnusi, line 2162: :heart_eyes:

gnusi (approved):

PR #280, opened 2021-08-11
fix seek for dense range iterator

gnusi:


core/formats/columnstore2.cpp
```diff
@@ -184,16 +186,21 @@ class range_column_iterator final
...
       std::get<document>(attrs_).value = doc;
       min_doc_ = doc + 1;
       std::get<irs::payload>(attrs_).value = this->payload(doc - min_base_);
       return doc;
     }
 
-    std::get<document>(attrs_).value = doc_limits::eof();
-    std::get<irs::payload>(attrs_).value = bytes_ref::NIL;
-    return doc_limits::eof();
+    if (doc != value()) {
```
Dronplane, line 196: And what if we seek backward?
E.g we have documents 1,2,3,4,5
So: min_doc_ = 1 max_doc_ = 5

seek(4)  will return 4 and set min_doc_ = 5.
What should we do in case of seek(1)  ?  With check here  we would set the iterator to eof. But I guess seek should return "the first reachable value of the iterator  greater or equal to the argument". So we should still return 4 here and do not move the iterator forward (or to eof).

PR #300, opened 2021-09-12
Get rid of `utf8_path` in favor of `std::filesystem::path`

gnusi:


tests/analysis/text_analyzer_tests.cpp
```diff
@@ -1002,12 +1000,10 @@ TEST_F(TextAnalyzerParserTestSuite, test_make_config_json) {
 
   // no stopwords, but empty stopwords path (we need to shift CWD to our test resources, to be able to load stopwords)
   {
-    auto oldCWD = irs::utf8_path(true);
-    auto newCWD = irs::utf8_path(IResearch_test_resource_dir);
-    newCWD.chdir();
-    auto reset_stopword_path = irs::make_finally([oldCWD]()noexcept{
-      oldCWD.chdir();
+    auto reset_stopword_path = irs::make_finally([oldCWD = std::filesystem::current_path()]()noexcept{
+      std::filesystem::current_path(oldCWD);
```
Dronplane, line 1004: Suggest to use noexcept version (with std::error_code out param) in the finally as previous version was also noexcept

Dronplane (approved):
LGTM. Only minor comment regarding noexcept

PR #301, opened 2021-09-14
bug-fix/tsan-issues-5

gnusi:


core/utils/object_pool.hpp
```diff
@@ -118,15 +114,16 @@ class concurrent_stack : private util::noncopyable {
 
       new_head.node = &new_node;
```
mpoeter, line 115: This line could be moved before the loop.

mpoeter (approved):

core/utils/object_pool.hpp
gnusi: good catch!

PR #302, opened 2021-09-23
Bug fix/restore utf8 path

Dronplane:
Temporary resurrect utf8_path class and use it instead of std::filesystem::path to support MacOS 10.14 where std::filesystem is not implemented.
Class interface was modified to match std::filesystem::path for simpler switch back once MacOs 10.14  support is dropped.


alexbakharew (approved):

alexbakharew (approved):

Dronplane:
Tests blue only known unrelated failure
https://jenkins.arangodb.biz:3456/view/PR/job/iresearch_s-pr/328/

PR #313, opened 2021-10-27
Bug fix/more tests columnstore

gnusi:
* honor primary sort for field features
* add more tests


core/index/field_data.cpp
```diff
@@ -1248,6 +1248,24 @@ void fields_data::flush_features(
...
+        std::begin(cached_features_),
+        std::end(cached_features_),
+        size_t{0},
+        [](const auto lhs, const auto& rhs) {
+          return lhs + rhs.stream.memory_active(); });
+}
+
+size_t fields_data::memory_reserved() const noexcept {
+  // FIXME(@gnusi): revisit the implementation
+  return byte_pool_.size() + int_pool_.size();
```
Dronplane, line 1266: Maybe ?
```suggestion
  return byte_pool_.size() + int_pool_.size() * sizeof(int_block_pool::value_type);
```

core/index/field_data.cpp
```diff
@@ -1099,9 +1113,9 @@ bool field_data::invert(token_stream& stream, doc_id_t id) {
     const auto res = terms_.emplace(term->value);
 
     if (nullptr == res.first) {
-      IR_FRMT_ERROR("field '%s' has invalid term of size '" IR_SIZE_T_SPECIFIER "'",
+      IR_FRMT_ERROR("field '%s' contains too long term of size '" IR_SIZE_T_SPECIFIER "'",
```
Dronplane, line 1116: Maybe lower this to "WARNING" ? As user will think something is broken if ERROR appears in log. And add "consequences" description like: "Term will be excluded from the index"

Dronplane (approved):
LGTM. Some minor comments

core/index/field_data.cpp
gnusi: already done in block_pool

PR #314, opened 2021-11-01
Hackup simdcomp build

markuspf:
This fixes building (for me) on Apple M1. 

I am really quite dubious that this is the correct fix for the problem: `simdcomp` seems to be a library for `x86_64` and it is unclear to me why it is built when on ARM.

### DIFF
--- a/external/simdcomp/CMakeLists.txt
+++ b/external/simdcomp/CMakeLists.txt
@@ -24,7 +24,7 @@ else()
 endif()
 
 # test for arm
-if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64.*|AARCH64.*)")
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64.*|AARCH64.*|arm64.*|aarch64.*)")
    set(BASE_FLAGS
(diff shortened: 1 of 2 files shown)


Dronplane:
https://jenkins.arangodb.biz:3456/view/iResearch/job/iresearch-release-s_aarch64/13/ Only expected tests failures

external/simdcomp/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ else()
 endif()
 
 # test for arm
-if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64.*|AARCH64.*)")
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64.*|AARCH64.*|arm64.*|aarch64.*)")
```
Dronplane, line 27: Maybe?
```suggestion
if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64.*|AARCH64.*|arm64.*|ARM64.*)")
```

Dronplane (changes requested):

PR #316, opened 2021-11-04
allow using CMPXCHG16

gnusi:
Adjust compiler definitions to allow using  `CMPXCHG16` instruction

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,8 +93,8 @@ option(SUPPRESS_EXTERNAL_WARNINGS "Suppress warnings originating in 3rd party co
 add_option_gprof(FALSE)
 
 # test for ARM
-if (CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64.*|AARCH64.*|arm64.*|ARM64.*)")
-  set(IS_ARM 1)
+if (CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64.*|AARCH64.*|arm64.*|ARM64.*|armv8.*|ARMV8.*)")
+  set(IS_ARM ON)
 endif()
 
 if (USE_VALGRIND)
@@ -123,7 +123,7 @@ if (MSVC)
     set(CMAKE_C_FLAGS "/MP ${CMAKE_C_FLAGS}")
     set(CMAKE_CXX_FLAGS "/MP ${CMAKE_CXX_FLAGS}")
   endif()
-elseif(NOT IS_ARM)
+elseif (NOT IS_ARM)
   # allow 16 byte compare-exchange
   add_compile_options(-mcx16)
 endif()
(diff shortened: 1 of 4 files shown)


gnusi:
@Dronplane do you to know how to enable `cmpxch16b` on msvc?

MBkkt:
> @Dronplane do you to know how to enable `cmpxch16b` on msvc?

https://github.com/microsoft/STL/commit/f3ca37d67c4bf18938497766a86aa78188b3ba3b#diff-06b7badfa74a470a7499cf1ca8451722f15f29ae3d6e119213a6550713cf05c0R381
https://github.com/microsoft/STL/commit/f3ca37d67c4bf18938497766a86aa78188b3ba3b#diff-880ef9f02349682be37ddb8618af2c0b4e5bfac857d94fa4e75f0ad224775bb5R14

https://docs.microsoft.com/en-us/cpp/intrinsics/interlockedcompareexchange128?view=msvc-160

Looks like it should be without any additional options, but only as intrinsic in C++17

CMakeLists.txt
```diff
@@ -92,6 +92,11 @@ option(SUPPRESS_EXTERNAL_WARNINGS "Suppress warnings originating in 3rd party co
 
 add_option_gprof(FALSE)
 
+# test for ARM
+if (CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64.*|AARCH64.*|arm64.*|ARM64.*)")
+  set(IS_ARM 1)
```
MBkkt, line 97:
```suggestion
if (CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64.*|AARCH64.*|arm64.*|ARM64.*|armv8.*)")
  set(IS_ARM ON)
```

CMakeLists.txt
```diff
@@ -114,10 +119,13 @@ if (MSVC)
   if (MSVC_BUILD_THREADS)
     set(CMAKE_C_FLAGS "/MP${MSVC_BUILD_THREADS} ${CMAKE_C_FLAGS}")
     set(CMAKE_CXX_FLAGS "/MP${MSVC_BUILD_THREADS} ${CMAKE_CXX_FLAGS}")
-  else ()
+  else()
     set(CMAKE_C_FLAGS "/MP ${CMAKE_C_FLAGS}")
     set(CMAKE_CXX_FLAGS "/MP ${CMAKE_CXX_FLAGS}")
   endif()
+elseif(NOT IS_ARM)
```
MBkkt, line 126: What about CMake format?

Dronplane:
Looks like 16byte  CAS  for Windows is only implemented for atomic_ref
Here is how defined the 16 byte atomic storage (note the reference!)
`template <class _Ty>
struct _Atomic_storage<_Ty&, 16> { // lock-free using 16-byte intrinsics`

Define _STD_ATOMIC_ALWAYS_USE_CMPXCHG16B could be used to force using _InterlockedCompareExchange128, but still that will not enable hardcoded "reference"

Dronplane (approved):
LGTM

PR #319, opened 2021-11-12
Add prebuilt for iresearch build

MBkkt:
Add prebuilt for iresearch build


prebuilt/external/snowball/libstemmer/modules.h
```diff
@@ -0,0 +1,265 @@
+/* /work/ArangoDB/build/3rdParty/iresearch/external/snowball/libstemmer/modules.h: List of stemming modules.
```
gnusi, line 1: I don't think we do need to use pre-built stuff from snowball as it doesn't bring any additional dependencies

Dronplane (approved):
LGTM.  MSVC  works as expected even with removed bison.

PR #334, opened 2022-01-17
Bug fix/enforce valid utf8

Dronplane:
By default ICU SortKey is just arbitrary bytes in range 0x00 - 0xFF
so in general SortKey is not valid UTF-8 sequence.
To achieve "validity" we split each byte in range 0x80 - 0xFF
in two bytes UTF-8 character. We don't care about "sanity" of this
UTF-8 string as SortKey was never intended to be human redable. And this
split still leaves the binary sorting order of modified SortKeys


core/analysis/collation_token_stream.cpp
```diff
@@ -269,26 +269,50 @@ bool collation_token_stream::reset(string_ref data) {
...
+      if ((termBufIdx + recalc.second) > sizeof state_->term_buf) {
+        IR_FRMT_ERROR(
+            "Collated token is more than %d bytes length after encoding.",
+            static_cast<int32_t>(sizeof state_->term_buf));
+        return false;
+      }
+      assert(recalc.second <= 2);
+      state_->term_buf[termBufIdx++] = kBytesRecalcMap[recalc.first];
+      if (recalc.second == 2) {
+        state_->term_buf[termBufIdx++] = kBytesRecalcMap[recalc.first + 1];
```
gnusi, line 309: it seems we don't need increment here

core/analysis/collation_token_stream.cpp
```diff
@@ -269,26 +269,50 @@ bool collation_token_stream::reset(string_ref data) {
...
+      term_size, static_cast<int32_t>(sizeof raw_term_buf));
     return false;
   }
+  size_t termBufIdx = static_cast<size_t>(term_size);
+  if (state_->options.forceUtf8) {
+    // enforce valid UTF-8 string
+    assert(buf == raw_term_buf);
+    termBufIdx = 0;
+    for (size_t i = 0; i < term_size; ++i) {
+      auto const& recalc = kRecalcMap[raw_term_buf[i]];
```
gnusi, line 299:
```suggestion
    const auto [offset, size] = kRecalcMap[raw_term_buf[i]];
```

gnusi, line 299: let's add an assertion to ensure we're within `kRecalcMap` boundaries

tests/analysis/collation_token_stream_test.cpp
```diff
@@ -30,10 +31,56 @@
...
   static_assert("collation" == irs::type<irs::analysis::collation_token_stream>::name());
 }
 
+TEST(collation_token_stream_test, test_byte_encoder) {
+  uint8_t target{0x0};
+  ASSERT_EQ(256, kRecalcMap.size());
+  do {
+    --target;
+    auto const expected = encode(target);
+    auto const actual = kRecalcMap[target];
```
gnusi, line 76: `const auto`

gnusi (changes requested):

PR #351, opened 2022-02-28
Feature/refactor directory

gnusi:
Refactor directory interface API to use std::span and std::string_view


core/analysis/analyzers.cpp
```diff
@@ -103,22 +103,22 @@ struct hash<::key> {
 
 namespace {
 
-const std::string FILENAME_PREFIX("libanalyzer-");
+const std::string kFileNamePrefix("libanalyzer-");
```
MBkkt, line 106: NIT
```suggestion
constexpr std::string_view kFileNamePrefix = "libanalyzer-";
```

core/formats/formats_burst_trie.cpp
```diff
@@ -1359,7 +1359,7 @@ void field_writer::prepare(const irs::flush_state& state) {
 }
 
 void field_writer::write(
-    const std::string& name,
+    std::string_view name,
```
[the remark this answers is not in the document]
MBkkt: [[maybe_unused]]

core/utils/directory_utils.hpp
```diff
@@ -40,13 +40,13 @@ namespace directory_utils {
...
   const directory& dir,
-  const std::string& name,
+  std::string_view name,
   bool include_missing = false);
 
 // return success, visitor gets passed references to files retrieved from source
 bool reference(
   const directory& dir,
-  const std::function<const std::string*()>& source,
+  const std::function<std::optional<std::string_view>()>& source,
```
MBkkt, line 49: Maybe old code better here?

core/utils/ref_counter.hpp
```diff
@@ -21,96 +21,107 @@
...
 #include "utils/noncopyable.hpp"
-#include "utils/thread_utils.hpp"
-#include "utils/memory.hpp"
 
 namespace iresearch {
 
-template<typename Key, typename Hash = absl::Hash<Key>, typename Equal = std::equal_to<Key>>
-class ref_counter : public util::noncopyable { // noncopyable because shared_ptr refs hold reference to internal map keys
+template<typename Key, typename Hash = absl::Hash<Key>,
+         typename Equal = std::equal_to<Key>>
```
MBkkt, line 38: Why not    
```
typename Hash = absl::container_internal::hash_default_hash<T>,
typename Equal = absl::container_internal::hash_default_eq<T>
```

MBkkt (approved):
Also, I don't like this code
```
std::string{some_string_view}.c_str()
```
But its more correct and it is executed if error, so not really important

core/analysis/analyzers.cpp
Dronplane: Code `std::string{some_string_view}.c_str()` is necessary as string_view is not guaranteed to be null terminated

MBkkt: I know, but I think in practice we use string_view constructed from string

gnusi: I hope we can improve that once `std::format(...)` is available.

core/utils/directory_utils.hpp
gnusi: There are places where I can't return a pointer to `std::string_view` because data stored as `std::string`. Anyways that's not heavy loaded place.

core/utils/ref_counter.hpp
gnusi: done

PR #361, opened 2022-04-26
Changes for CMakeLists.txt location

alexbakharew:


MBkkt (commented):
lgtm

cmake/FindLz4.cmake
```diff
@@ -71,9 +71,25 @@ find_path(Lz4_SRC_DIR_LZ4
 if (Lz4_SRC_DIR_LZ4)
   get_filename_component(Lz4_SRC_DIR_PARENT ${Lz4_SRC_DIR_LZ4} DIRECTORY)
 
+  set(LZ4_CMAKE_PATH "")
```
Dronplane, line 74: As a wild idea - what if check here if `LZ4_CMAKE_PATH` is empty sting? If it is not (e.g. someone set it via Cmake parameters) just use provided value. And only if it is not set - do the search?

alexbakharew: AFAIK we don't set LZ4_CMAKE_PATH previously. I am not sure it is really useful

PR #364, opened 2022-05-12
Make skip list aware of scorer

gnusi:


gnusi:
Closed in favor of https://github.com/iresearch-toolkit/iresearch/pull/394

PR #365, opened 2022-05-13
Nested search preparation

gnusi:
Ensure document blocks are not splitted during consolidation

### DIFF
--- a/core/index/merge_writer.cpp
+++ b/core/index/merge_writer.cpp
@@ -1092,12 +1092,12 @@ class SortingCompoundDocIterator : util::noncopyable {
       assert(lhs < itrs_.size());
       assert(rhs < itrs_.size());
 
-      const auto& lhs_it = itrs_[lhs];
-      const auto& rhs_it = itrs_[rhs];
+      const auto& [lhs_it, lhs_doc, lhs_pay] = itrs_[lhs];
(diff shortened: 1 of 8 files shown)


core/index/merge_writer.cpp
```diff
@@ -1109,103 +1044,96 @@ std::optional<field_id> columnstore::insert(
...
+      } else if ((*less_)(lhs_value, rhs_value)) {
+        return false;
+      }
+
+      // tie braker to avoid splitting document blocks
+      if (lhs_it.it.get() == rhs_it.it.get()) {
+        return lhs_it.doc->value < rhs_it.doc->value;
+      }
+
+      return lhs_it.it.get() < rhs_it.it.get();
```
MBkkt, line 1112: Why not?
```suggestion
      return lhs_it.it.get() < rhs_it.it.get() || 
             (lhs_it.it.get() == rhs_it.it.get() && lhs_it.doc->value < rhs_it.doc->value);
```

MBkkt (approved):
LGTM

Dronplane (approved):
LGTM

core/index/merge_writer.cpp
gnusi: Isn't it the same? ;-)

PR #367, opened 2022-05-17
Flushed segments & Death tests

alexbakharew:
- Problem with losing segments after error during flush is fixed
- New death tests for columnstore 1.4

### DIFF
--- a/tests/index/index_death_tests.cpp
+++ b/tests/index/index_death_tests.cpp
@@ -1,4 +1,4 @@
-////////////////////////////////////////////////////////////////////////////////
+﻿////////////////////////////////////////////////////////////////////////////////
 /// DISCLAIMER
 ///
 /// Copyright 2018 ArangoDB GmbH, Cologne, Germany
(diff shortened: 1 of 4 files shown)


tests/index/index_death_tests.cpp
```diff
@@ -25,6 +25,8 @@
 #include "index_tests.hpp"
 #include "formats/formats.hpp"
 #include "store/memory_directory.hpp"
+#include "store/mmap_directory.hpp"
+#include "store/fs_directory.hpp"
```
alexbakharew, line 29: Should be deleted

tests/index/index_death_tests.cpp
```diff
@@ -2716,26 +2722,26 @@ TEST(index_death_test_formats_14, columnstore_creation_fail_implicit_segment_flu
...
+
     ASSERT_THROW(insert(*writer,
       doc2->indexed.begin(), doc2->indexed.end(),
       doc2->stored.begin(), doc2->stored.end()), irs::io_error);
 
-    ASSERT_FALSE(writer->commit());
+    dir.length(length, "_3.csd");
+    ASSERT_GT(length, 0);
+
+//    ASSERT_FALSE(writer->commit());
```
alexbakharew, line 2735: Should be deleted

alexbakharew (commented):
clean up code

alexbakharew:
Jenkins seems to be broken right now

PR #371, opened 2022-06-15
Adjust nested query logic for border cases

gnusi:


core/search/nested_filter.cpp
```diff
@@ -65,6 +65,33 @@ bool IsValid(const ByNestedOptions::MatchType& match) noexcept {
...
 
+class ScorerWrapper final : public doc_iterator {
+ public:
+  explicit ScorerWrapper(doc_iterator::ptr it, ScoreFunction&& score) noexcept
+      : it_{std::move(it)} {
+    assert(it_);
+    score_ = std::move(score);
+  }
+
+  doc_id_t value() const override { return it_->value(); }
```
MBkkt, line 76: NIT
I suggest final because the class may become non-final sometime

core/search/nested_filter.cpp
```diff
@@ -571,6 +632,41 @@ doc_iterator::ptr ByNesterQuery::execute(const sub_reader& rdr,
...
+
+                    if (auto* scr = irs::get_mutable<score>(parent.get());
+                        scr) {
+                      *scr = std::move(func);
+                      return std::move(parent);
+                    } else {
+                      return memory::make_managed<ScorerWrapper>(
+                          std::move(parent), std::move(func));
+                    }
+                  }
```
MBkkt, line 651: NIT
```suggestion
                  if constexpr (!std::is_same_v<NoopAggregator, A>) {
                    auto func = ScoreFunction::Constant(none_boost_,
                                                        ord.buckets().size());
                    auto* scr = irs::get_mutable<score>(parent.get());
                    if (!scr) {
                      return memory::make_managed<ScorerWrapper>(
                          std::move(parent), std::move(func));                      
                    }
                    *scr = std::move(func);
                  }
                  return std::move(parent);
```

core/search/nested_filter.cpp
```diff
@@ -437,22 +499,19 @@ class MinMatcher : public Merger,
...
     const auto& child_score = *self.child_score_;
 
-    doc_id_t count = min_;
-
     if constexpr (HasScore_v<Merger>) {
       child_score(buf.data());
     }
+
     while (child.next() && child_doc->value < parent) {
+      if (!--count) {
```
MBkkt, line 507:
```suggestion
      if (--count == 0) {
```

core/search/nested_filter.cpp
```diff
@@ -425,10 +470,27 @@ class MinMatcher : public Merger,
...
+      return 0;
+    }
+
     if (first_child > parent) {
       return first_child;
     }
 
+    doc_id_t count = min_ - 1;
+
+    if (!count) {
```
MBkkt, line 490:
```suggestion
    if (count == 0) {
```

MBkkt, line 490: Can you explain to me why we cannot check it before `first_child > parent` check?

Dronplane (approved):
LGTM

MBkkt (approved):

PR #377, opened 2022-06-27
Fix build for MSVC 2022

Dronplane:
__FUNCSIG__ is not detected by "ifdef" but still works.
/Zc:static_assert- flag added to disable  breaking change ins static assert evaluation  see  https://docs.microsoft.com/en-us/cpp/overview/cpp-conformance-improvements?view=msvc-170#error-on-a-non-dependent-static_assert


MBkkt:
#ifndef FUNC_NAME
#  if defined(__clang__) || defined(__GNUC__)
#    define FUNC_NAME __PRETTY_FUNCTION__
#  elif defined(_MSC_VER) && some version if needed
#    define FUNC_NAME __FUNCSIG__
#  else
#    define FUNC_NAME __func__
#  endif
#endif

MBkkt (approved):

PR #388, opened 2022-08-24
Workaround MSVC compiler stack overflow.

mpoeter:
New version of VS2022 (17.3.2; cl.exe version 19.33.31629.0) ran into a stack overflow when building normalization_data_cp_props.cpp. This workaround avoids that. I will also create a bug report with MS.


Dronplane (approved):
LGTM

PR #391, opened 2022-08-31
Harden some checks.

gnusi:


core/search/min_match_disjunction.hpp
```diff
@@ -159,18 +159,12 @@ class min_match_disjunction : public doc_iterator,
...
 
-    /* execute seek for all lead iterators and
-     * move one to head if it doesn't hit the target */
+    // execute seek for all lead iterators and
+    // move one to head if it doesn't hit the target
     for (auto it = lead(), end = heap_.end(); it != end;) {
       assert(*it < itrs_.size());
       const auto doc = itrs_[*it]->seek(target);
 
-      if (doc == target) {
```
MBkkt, line 168: This code change about check target != eof?

MBkkt: Why not check it once?

Dronplane (approved):
LGTM

core/search/min_match_disjunction.hpp
gnusi: no, this checks if doc is equal to target after seek(...). The check was moved.

PR #397, opened 2022-09-12
Fix handling of offsets for short ngram sequences

gnusi:


core/search/ngram_similarity_query.cpp
```diff
@@ -411,6 +415,11 @@ bool SerialPositionsChecker<Base>::Check(size_t potential, doc_id_t doc) {
       }
     } else {
       freq = 1;
+      if constexpr (HasPosition) {
+        if (last_state) {
```
Dronplane, line 419: Should we just assert `last_state` and add ARS_LIKELY to condition?

Dronplane (approved):
LGTM

MBkkt (approved):

PR #400, opened 2022-09-23
Allow specifying all docs provider for boolean queries

gnusi:


MBkkt (approved):

core/search/boolean_filter.cpp
```diff
@@ -477,15 +488,16 @@ filter::prepared::ptr Or::prepare(std::vector<const filter*>& incl,
...
+  const auto all_type = cumulative_all->type();
   size_t optimized_match_count = 0;
   // Optimization steps
 
   score_t all_boost{0};
   size_t all_count{0};
   const irs::filter* incl_all{nullptr};
   for (auto filter : incl) {
-    if (filter->type() == irs::type<irs::all>::id()) {
+    if (filter->type() == all_type) {
```
Dronplane, line 500: Just type check is not enough. As filter used as "all" could match type of other filters used. And that filters would be collapsed here

core/search/boolean_filter.cpp
```diff
@@ -400,21 +409,22 @@ filter::prepared::ptr And::prepare(std::vector<const filter*>& incl,
...
   }
 
-  irs::all cumulative_all;
+  auto cumulative_all = all_docs(1.f);
+  const auto all_type = cumulative_all->type();
   score_t all_boost{0};
   size_t all_count{0};
   for (auto filter : incl) {
-    if (filter->type() == irs::type<irs::all>::id()) {
+    if (filter->type() == all_type) {
```
Dronplane, line 417: Just type check is not enough. Filter should completely match "custom" all

Dronplane (approved):
LGTM

MBkkt (approved):

PR #403, opened 2022-10-14
Refactor string ref

gnusi:


core/utils/string.hpp
```diff
@@ -375,143 +172,66 @@ inline size_t common_prefix_length(const Char* lhs, size_t lhs_size,
...
-  const basic_string_ref<_Elem, _Traits>& first,
-  const basic_string_ref<_Elem, _Traits>& second) {
-  return starts_with(first, second.c_str(), second.size());
-}
-
-typedef basic_string_ref<char> string_ref;
-typedef basic_string_ref<byte_type> bytes_ref;
+template<typename Char>
+inline std::basic_string_view<Char> kEmptyStringView{
+  reinterpret_cast<const Char*>(""), 0};
```
MBkkt, line 183: `L""` not same in bytes as `""`

Probably will be ok something like:
```cpp
inline constexpr char const* nulls = "\0\0\0";
```
https://godbolt.org/z/6xce3TYoE

MBkkt (commented):
Don't see other files

Dronplane (approved):
LGTM

PR #405, opened 2022-10-23
Fix finally

MBkkt:


gnusi (changes requested):
Please add `fu2` to https://github.com/iresearch-toolkit/iresearch/blob/master/THIRD_PARTY_README.md

core/utils/misc.hpp
```diff
@@ -33,57 +32,37 @@ namespace iresearch {
...
 // Convenient helper for simulating 'try/catch/finally' semantic
 template<typename Func>
-class finally {
+class [[nodiscard]] Finally {
  public:
   static_assert(std::is_nothrow_invocable_v<Func>);
 
-  explicit finally(Func&& func) : func_{std::forward<Func>(func)} {}
-  ~finally() { func_(); }
+  // If you need some of it, please use absl::MakeCleanup
```
Dronplane, line 39: Why not completely move to absl::Cleanup ?

Dronplane (approved):
LGTM

core/utils/misc.hpp
MBkkt: We discussed it earlier with @gnusi and for simple expression we want simple (without flag, placement new, etc) abstraction.

But copy/move finally is unsafe (func will be called twice or more time)

gnusi (approved):

PR #409, opened 2022-10-31
Bug fix/fix tick on commit

Dronplane:
Adds explicit tick for documents_context and fix generations

### DIFF
--- a/tests/index/index_tests.cpp
+++ b/tests/index/index_tests.cpp
@@ -6259,91 +6259,70 @@ TEST_P(index_test_case, refresh_reader) {
   }
 
   // validate state pre/post refresh (existing segment changed)
-  {{ASSERT_EQ(1, reader.size());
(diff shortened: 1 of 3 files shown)


tests/index/index_tests.cpp
```diff
@@ -16401,6 +16403,141 @@ TEST_P(index_test_case_11, commit_payload) {
...
+      ASSERT_TRUE(doc);
+    }
+    // subcontext with remove
+    {
+      auto trx2 = writer->documents();
+      trx2.SetFirstTick(1);
+      trx2.remove(MakeByTerm("name", "A"));
+      trx2.SetLastTick(2);
+    }
+    trx.SetLastTick(2);
```
MBkkt, line 16530: It's not possible

MBkkt (commented):
What about case remove before insert?

first also cannot be equal, because first == (last - count primitive operations (insert + remove + 2 * update))

MBkkt (approved):

PR #434, opened 2022-11-22
Feature/directory caching

gnusi:


core/store/caching_directory.hpp
```diff
@@ -0,0 +1,146 @@
...
+    if (!stream) {
+      return nullptr;
+    }
+
+    {
+      std::lock_guard lock{mutex_};
+      if (Acceptor::operator()(cache_.size(), name, advice)) {
+        try {
+          const auto [it, _] = cache_.try_emplace(name, std::move(stream));
+          return it->second->reopen();
```
MBkkt, line 123: same here

core/store/caching_directory.hpp
```diff
@@ -0,0 +1,146 @@
...
+        } catch (...) {
+        }
+      }
+      return true;
+    }
+    return false;
+  }
+
+  index_input::ptr open(std::string_view name,
+                        IOAdvice advice) const noexcept override {
```
MBkkt, line 103: I think better to write separate OpenImpl function which return index_input::ptr and here you just call reopen on it (if not null)

core/store/caching_directory.hpp
```diff
@@ -0,0 +1,146 @@
...
+  bool rename(std::string_view src, std::string_view dst) noexcept override {
+    if (Impl::rename(src, dst)) {
+      std::lock_guard lock{mutex_};
+      if (auto src_it = cache_.find(src); src_it != cache_.end()) {
+        auto tmp = std::move(src_it->second);
+        cache_.erase(src_it);
+        try {
+          cache_[dst] = std::move(tmp);
+        } catch (...) {
+        }
```
MBkkt, line 95: it's not ok if we ignore exception here, I think if we don't have enough memory in this situation it's ok to terminate

core/store/mmap_directory.hpp
```diff
@@ -32,10 +31,7 @@ class mmap_directory : public fs_directory {
   explicit mmap_directory(std::filesystem::path dir,
                           directory_attributes attrs = directory_attributes{});
 
-  virtual index_input::ptr open(std::string_view name,
-                                IOAdvice advice) const noexcept override final;
+  index_input::ptr open(std::string_view name, IOAdvice advice) const noexcept;
```
MBkkt, line 34: is it virtual? override?

core/store/caching_directory.hpp
```diff
@@ -0,0 +1,146 @@
...
+};
+
+template<typename Impl, typename Acceptor = MaxSizeAcceptor>
+class CachingDirectory : public Impl, private Acceptor {
+ public:
+  template<typename... Args>
+  explicit CachingDirectory(const Acceptor& acceptor, Args&&... args)
+    : Impl{std::forward<Args>(args)...}, Acceptor{acceptor} {}
+
+  bool exists(bool& result, std::string_view name) const noexcept override {
```
MBkkt, line 56: I think all member function, at least now can be final

core/store/caching_directory.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+    auto stream = Impl::open(name, advice);
+
+    if (!stream) {
+      return nullptr;
+    }
+
+    {
+      std::lock_guard lock{mutex_};
+      if (Acceptor::operator()(cache_.size(), name, advice)) {
```
MBkkt, line 120: what is acceptor?
What if two threads create "same" stream, wait mutex,
For first thread acceptor return true -> make reopen
For second false -> return stream?

core/store/caching_directory.hpp
```diff
@@ -0,0 +1,146 @@
...
+        try {
+          const auto [it, _] = cache_.try_emplace(name, std::move(stream));
+          return it->second->reopen();
+        } catch (...) {
+        }
+      }
+    }
+
+    return stream;
```
MBkkt, line 129: why?

in general I expected something like

```cpp
compute hash;
if (shared_lock lock{mutex}; auto input = get(key, hash)) {
  lock.unlock();
  return input.reopen();
}
auto input = Impl::open(name, advice);
unique_lock lock{mutex};
const auto [it, was_insert] = cache_.try_emplace(name, std::move(input));
if (was_insert) {
  lock.unlock();
  return input;
} else {
  auto old_input = it->second;
  lock.unlock();
  input.close/remove();
  return old_input->reopen();
}
```

core/store/caching_directory.hpp
MBkkt, line 123: You call reopen for file which was open few lines ago

MBkkt (changes requested):

core/store/caching_directory.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+#include <shared_mutex>
+
+#include "store/directory.hpp"
+
+namespace iresearch {
+
+class MaxSizeAcceptor {
+ public:
+  explicit constexpr MaxSizeAcceptor(size_t max_size = 1024) noexcept
```
Dronplane, line 36: NIT 
MaxCountAcceptor would be better. As it is count of files, not overall size

core/store/caching_directory.hpp
```diff
@@ -0,0 +1,146 @@
...
+        try {
+          cache_[dst] = std::move(tmp);
+        } catch (...) {
+        }
+      }
+      return true;
+    }
+    return false;
+  }
+
```
MBkkt, line 101:
```cpp
bool find(key, Visitor&& v) {
      hash = std::hash<>{}(key);
      std::shared_lock lock{mutex_};
      if (auto it = cache_.find(name, hash); it != cache_.end()) {
          v(it);
          return true;
      }
      return false;
}
```

Dronplane (approved):
LGTM

core/store/caching_directory.hpp
```diff
@@ -0,0 +1,161 @@
...
+};
+
+template<typename Impl, typename Acceptor = MaxCountAcceptor>
+class CachingDirectory : public Impl, private Acceptor {
+ public:
+  template<typename... Args>
+  explicit CachingDirectory(const Acceptor& acceptor, Args&&... args)
+    : Impl{std::forward<Args>(args)...}, Acceptor{acceptor} {}
+
+  bool exists(bool& result, std::string_view name) const noexcept override {
```
MBkkt, line 56: final?

MBkkt (approved):

core/store/caching_directory.hpp
```diff
@@ -0,0 +1,146 @@
+
```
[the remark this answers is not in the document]
MBkkt: You forgot to remove it

PR #438, opened 2022-11-24
Cleanup/c++20 ebo

gnusi:


tests/utils/container_utils_tests.cpp
```diff
@@ -42,12 +42,8 @@ TEST(container_utils_tests, test_bucket_allocator) {
...
   irs::container_utils::memory::bucket_allocator<bucket_builder, 16> alloc(1);
 
-  typedef irs::container_utils::raw_block_vector<16,  // total number of levels
-                                                 8,   // size of the first block
-                                                      // 2^8
-                                                 decltype(alloc)  // allocator
-                                                 >
-    raw_block_vector_t;
+  using raw_block_vector_t =
+    irs::container_utils::raw_block_vector<16, 8, decltype(alloc)>;
```
MBkkt, line 46: Why?

core/utils/misc.hpp
```diff
@@ -48,7 +79,7 @@ class [[nodiscard]] Finally {
   ~Finally() noexcept { func_(); }
 
  private:
-  Func func_;
+  IRS_NO_UNIQUE_ADDRESS Func func_;
```
MBkkt, line 82: Nice

core/utils/misc.hpp
```diff
@@ -31,6 +31,37 @@
 
 namespace iresearch {
 
+template<typename T>
+class EboRef {
```
MBkkt, line 35: Is it really necessary?
I think it temporary solution to avoid changes a lot of code?
Can you write a comment about that?

Also I think it will be better in separate header, for example old ebo header

MBkkt (approved):

Dronplane (approved):
LGTM

tests/utils/container_utils_tests.cpp
gnusi: why what? ;-)

MBkkt: Why do you remove that comments

gnusi: It breaks formatting

PR #440, opened 2022-11-25
Add check to avoid warning generation

Dronplane:

PR #441, opened 2022-11-25
Remove optimization based on not possibility ub

MBkkt:


Dronplane (approved):
LGTM

Dronplane:
Alternative solution for https://github.com/iresearch-toolkit/iresearch/pull/440

MBkkt:
We decide smaller solution is better suited

PR #445, opened 2022-11-29
Fix compilation in ArangoDB

MBkkt:
and cleanup boost includes

### DIFF
--- a/core/search/boolean_filter.cpp
+++ b/core/search/boolean_filter.cpp
@@ -22,7 +22,7 @@
 
 #include "boolean_filter.hpp"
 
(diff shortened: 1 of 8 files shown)


core/search/boolean_filter.cpp
```diff
@@ -22,7 +22,7 @@
 
 #include "boolean_filter.hpp"
 
-#include <boost/functional/hash.hpp>
```
MBkkt, line 25: https://www.boost.org/doc/libs/1_80_0/boost/functional/hash/hash.hpp

PR #446, opened 2022-11-29
Update abseil to fix compile error in arangodb

MBkkt:


MBkkt:
https://jenkins01.arangodb.biz/job/iresearch-release/1921/

PR #450, opened 2022-12-02
Add lightweight accessor to term's documents

gnusi:


core/formats/formats_burst_trie.cpp
```diff
@@ -3121,8 +3112,54 @@ class field_reader final : public irs::field_reader {
...
+      }
+
+      auto docs_it = it.postings(IndexFeatures::NONE);
+
+      const auto* doc = irs::get<document>(*docs_it);
+
+      if (IRS_UNLIKELY(!doc)) {
+        assert(false);
+        return 0;
+      }
```
MBkkt, line 3149: I don't like it, why do we should write some code which never executed.

core/formats/formats_burst_trie.cpp
```diff
@@ -3121,8 +3112,54 @@ class field_reader final : public irs::field_reader {
...
+
+      auto* p = docs;
+
+      while (count && docs_it->next()) {
+        *p++ = doc->value;
+        --count;
+      }
+
+      return std::distance(docs, p);
```
MBkkt, line 3158: NIT
instead of this you can use for `(; iter < count && docs_it->next(); ++iter)`

core/formats/formats_burst_trie.cpp
```diff
@@ -3121,8 +3112,54 @@ class field_reader final : public irs::field_reader {
...
+      single_term_iterator<FST> it{meta(), *owner_->pr_,
+                                   owner_->terms_in_->reopen(),
+                                   owner_->terms_in_cipher_.get(), *fst_};
+
+      it.seek(term);
+      return it.meta();
+    }
+
+    size_t read_documents(bytes_view term, doc_id_t* docs,
+                          size_t count) const override {
```
MBkkt, line 3125: Why not?
```suggestion
    size_t read_documents(bytes_view term, std::span<doc_id_t> docs) const override {
```

MBkkt (approved):

Dronplane (approved):
LGTM. But current implementation also returns deleted document. That might be not desired.

PR #451, opened 2022-12-06
Add IRS_ASSERT

MBkkt:


core/utils/type_utils.hpp
```diff
@@ -37,7 +37,7 @@ namespace iresearch {
 ////////////////////////////////////////////////////////////////////////////////
 template<typename T>
 constexpr std::string_view ctti() noexcept {
-  return {IRESEARCH_CURRENT_FUNCTION};
+  return {IRS_FUNC_NAME};
```
MBkkt, line 40: .

core/utils/numeric_utils.cpp
```diff
@@ -142,7 +142,7 @@ typename EncodeTraits::type decode(const byte_type* in) {
   return value;
 }
 
-// assert that signed right shift works as expected
+// IRS_ASSERT that signed right shift works as expected
```
MBkkt, line 145: .

core/utils/assert.hpp
```diff
@@ -0,0 +1,77 @@
...
+#pragma once
+
+#include <cstddef>
+#include <string_view>
+
+#include "shared.hpp"
+
+namespace iresearch {
+
+enum class LogLevel : unsigned char {
```
Dronplane, line 32: Why not use log levels/log callback from log.hpp/cpp ? We could just add here additional level for assertion if needed (or just use IRL_FATAL)

MBkkt: I don't want add new global value

Dronplane: Then why  enum? Are we expecting here many asserion log levels?

core/utils/assert.cpp
```diff
@@ -0,0 +1,48 @@
...
+  if (const auto callback = sCallbacks[static_cast<std::size_t>(level)];
+      IRS_LIKELY(callback != nullptr)) {
+    callback(file, line, func, condition);
+  }
+}
+
+}  // namespace detail
+
+void SetCallback(LogLevel level, LogCallback callback) noexcept {
+  detail::sCallbacks[static_cast<std::size_t>(level)] = callback;
```
Dronplane, line 45: Suggest validating `level` value here. As this value comes from outside of the library. If we decide in the end to have multiple assertion levels

MBkkt: It's enum class, why it should be validated?

gnusi:
Please set callback in `iresearch/utils/main.cpp`  as well

tests/tests_main.cpp
```diff
@@ -362,11 +362,15 @@ void test_base::SetUp() {
...
 }
 
-// -----------------------------------------------------------------------------
-// --SECTION--                                                              main
-// -----------------------------------------------------------------------------
+static void AssertCallback(std::string_view file, std::size_t line,
+                           std::string_view function,
+                           std::string_view condition) noexcept {
+  FAIL() << file << ":" << line << ": " << function << ": Condition '"
+         << condition << "' is true.";
```
gnusi, line 369: maybe `is false`?

core/formats/formats_10.cpp
```diff
@@ -2904,8 +2892,9 @@ void segment_meta_writer::write(directory& dir, std::string& meta_file,
 }
 
 struct segment_meta_reader final : public irs::segment_meta_reader {
-  virtual void read(const directory& dir, segment_meta& meta,
-                    std::string_view filename = {}) override;  // null == use meta
+  virtual void read(
```
gnusi, line 2895: remove virtual

core/formats/formats_10.cpp
```diff
@@ -2629,9 +2616,10 @@ struct index_meta_reader final : public irs::index_meta_reader {
   virtual bool last_segments_file(const directory& dir,
                                   std::string& name) const override;
 
-  virtual void read(const directory& dir, index_meta& meta,
-                    std::string_view filename = {}) override;  // null == use meta
-};                                                       // index_meta_reader
+  virtual void read(
```
gnusi, line 2619: remove virtual

core/shared.hpp
```diff
@@ -102,118 +60,41 @@
...
-#define GCC_ONLY(...) __VA_ARGS__
-#else
-#define GCC_ONLY(...)
-#endif
-
 // check if sizeof(float_t) == sizeof(double_t)
 #if defined(FLT_EVAL_METHOD) && \
   ((FLT_EVAL_METHOD == 1) || (FLT_EVAL_METHOD == 2))
 static_assert(sizeof(float_t) == sizeof(double_t),
               "sizeof(float_t) != sizeof(double_t)");
```
gnusi, line 67: also remove message?

core/shared.hpp
```diff
@@ -230,22 +111,6 @@ static_assert(sizeof(float_t) != sizeof(double_t),
 #define IRESEARCH_SSE2
 #endif
 
-#ifdef __SSE4_1__
```
gnusi, line 233: I prefer to keep it for future

core/shared.hpp
```diff
@@ -102,118 +60,41 @@
...
-#ifndef __has_attribute
-#define IRESEARCH_HAS_ATTRIBUTE(x) 0
-#else
-#define IRESEARCH_HAS_ATTRIBUTE(x) __has_attribute(x)
-#endif
-
-// IRESEARCH_ATTRIBUTE_NONNULL
-#if IRESEARCH_HAS_ATTRIBUTE(nonnull) || \
-  (defined(__GNUC__) && !defined(__clang__))
+#if __has_attribute(nonnull)
```
gnusi, line 97: what if __has_attribute isn't defined?

core/shared.hpp
```diff
@@ -102,118 +60,41 @@
...
-#define IRESEARCH_CURRENT_FUNCTION __PRETTY_FUNCTION__
+#define IRS_FUNC_NAME __FUNCSIG__
 #else
-#error "compiler is not supported"
+#define IRS_FUNC_NAME __func__
 #endif
-
-// IRESEARCH_COMPILER_HAS_FEATURE
-#ifndef __has_feature
-#define IRESEARCH_COMPILER_HAS_FEATURE(x) \
```
gnusi, line 201: let's keep it

core/shared.hpp
```diff
@@ -102,118 +60,41 @@
...
-#define IRESEARCH_PLUGIN_EXPORT extern "C" IRESEARCH_HELPER_DLL_EXPORT
-#else
-#define IRESEARCH_PLUGIN_EXPORT
-#endif  // IRESEARCH_DLL_PLUGIN
-#else   // IRESEARCH_DLL is not defined: this means IRESEARCH is a static lib.
-#define IRESEARCH_API
-#endif  // IRESEARCH_DLL
-
 // MSVC 2015 does not define __cpp_lib_generic_associative_lookup macro
 #if (defined(__cpp_lib_generic_associative_lookup) || \
```
gnusi, line 82: I don't think we need it

core/shared.hpp
MBkkt: It always defined in supported compilers

core/shared.hpp
MBkkt: Why?

core/shared.hpp
MBkkt: Ok

core/utils/assert.hpp
MBkkt: I think maybe in future we remove log with printf format style

tests/tests_main.cpp
MBkkt: No, true, because LOG_MESSAGE(condition) condition should be true
assert add `!`

core/shared.hpp
MBkkt: Anyway done

core/shared.hpp
MBkkt: I don't think it's necessary, but ok

core/utils/assert.hpp
MBkkt:
```cpp
#define LOG_ERROR(cond, message)
 do {
   if (cond) {
     LogMessage(source_location, #cond, message)
   }
 } while(false)


// message will be computed only if condition true
LOG_ERROR(result.isError(), absl::StrCat/Substitute(...)); 
```

Dronplane: Still don't get it, why we need separate log levels enum for assertions, with just single value in it. If we will replace printf style logger (I'm totally for it) we will add proper log levele there.

tests/tests_main.cpp
Dronplane: Then we could supply condition as string as separate macros arg and avoid this  "!<condition> is true"
Like this:
#define ASSERT_PRINTER(cond, cond_text)                           \
  do {                                                           \
    if (!(cond)) {                                              \
      std::cout <<  __FILE__ << ", " << __LINE__ << "[" << cond_text << "] is false" << std::endl;     \
    }                                                            \
  } while (false)

#define IRS_ASSERT(cond) ASSERT_PRINTER(cond, #cond)

MBkkt: I don't think formatting in macro is good idea, anyway fixed

core/utils/assert.hpp
```diff
@@ -0,0 +1,99 @@
...
+  do {                                                                      \
+    if (!(cond)) {                                                          \
+      ::iresearch::detail::AssertMessage(__FILE__, __LINE__, IRS_FUNC_NAME, \
+                                         #cond, message);                   \
+    }                                                                       \
+  } while (false)
+
+#define IRS_GET_MACRO(arg1, arg2, MACRO, ...) MACRO
+
+#ifdef NDEBUG
```
Dronplane, line 61: Should this also be controlled by `IRESEARCH_DEBUG` ?

MBkkt: No

MBkkt: This just remove unnecessary if (false) etc in release

Dronplane: But RelWithDebugInfo also defines `NDEBUG`.
I was expecting that whole assertions should either trigger or not based only on `IRESEARCH_DEBUG`

MBkkt: >But RelWithDebugInfo also defines NDEBUG.

And it's good

MBkkt: > I was expecting that whole assertions should either trigger or not based only on IRESEARCH_DEBUG

Yes, NDEBUG defines only IRS_STUB implementation

Dronplane (approved):
LGTM

PR #457, opened 2022-12-09
Bug fix/mac

gnusi:


core/index/sorted_column.cpp
```diff
@@ -72,8 +72,15 @@ bool sorted_column::flush_sprase_primary(
     ++new_doc;
   }
 
+  // Ensure that all docs up to new_doc are remapped without gaps
+  IRS_ASSERT(std::all_of(docmap.begin(), docmap.begin() + new_doc,
+                         [](doc_id_t doc) { return !doc_limits::eof(doc); }));
```
MBkkt, line 77:
```suggestion
                         [](doc_id_t doc) { return doc_limits::valid(doc); }));
```

tests/index/assert_format.cpp
```diff
@@ -261,11 +261,15 @@ void index_segment::compute_features() {
...
-  irs::bytes_output out{buf_};
-  if (f && f->write(out)) {
-    sort_.emplace_back(std::move(buf_), doc(), empty_count_);
-    empty_count_ = 0;
+  if (f) {
+    buf_.clear();
+    irs::bytes_output out{buf_};
+    if (f->write(out)) {
+      sort_.emplace_back(std::move(buf_), doc(), empty_count_);
+      empty_count_ = 0;
```
MBkkt, line 269: early return instead of duplicate else code block?

core/formats/skip_list.cpp
```diff
@@ -45,8 +45,7 @@ namespace irs {
 void SkipWriter::Prepare(
   size_t max_levels, size_t count,
   const memory_allocator& alloc /* = memory_allocator::global() */) {
-  max_levels_ =
-    std::clamp(max_levels, size_t{1}, ::max_levels(skip_0_, skip_n_, count));
```
MBkkt, line 49: Why is it UB?

gnusi: `::max_levels(skip_0_, skip_n_, count)`  can return 0

MBkkt: 👍

PR #459, opened 2022-12-12
Various fixes

gnusi:
* Handle nonexistent terms earlier
* Ensure block_disjunction accepts already exhausted iterators
* Fix for seek to the cached term


Dronplane (approved):
LGTM

PR #467, opened 2022-12-20
Fix assert which failed in ArangoDB

MBkkt:
in AddToPending (which can be called before commit) we can copy segment pointer, so we should increment use_count variabe 

TODO Remove all this useless variable or move they under IRESEARCH_DEBUG

### DIFF
--- a/core/index/index_writer.hpp
+++ b/core/index/index_writer.hpp
@@ -314,7 +314,7 @@ class index_writer : private util::noncopyable {
     // the segment_context used for storing changes (lazy-initialized)
     active_segment_context segment_;
     // segment_.ctx().use_count() at constructor/destructor time must equal
-    uint64_t segment_use_count_{0};
+    long segment_use_count_{0};
     uint64_t last_operation_tick_{0};   // transaction commit tick
     uint64_t first_operation_tick_{0};  // transaction tick
     index_writer& writer_;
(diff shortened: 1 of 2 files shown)


core/index/index_writer.hpp
```diff
@@ -314,7 +314,7 @@ class index_writer : private util::noncopyable {
     // the segment_context used for storing changes (lazy-initialized)
     active_segment_context segment_;
     // segment_.ctx().use_count() at constructor/destructor time must equal
-    uint64_t segment_use_count_{0};
+    long segment_use_count_{0};
```
Dronplane, line 317: Why signed type? Do we expect negative values?

Dronplane (commented):
Is i possible to make a test to reproduce the issue here?

core/index/index_writer.hpp
MBkkt: No, but use_count return long, I don't want to make cast
https://stackoverflow.com/questions/36370180/why-does-shared-ptrtuse-count-return-a-long-instead-of-an-unsigned-type

MBkkt: > Is i possible to make a test to reproduce the issue here?

Maybe... But I still want to remove this checks, so I don't think we need it

PR #469, opened 2022-12-27
prepare iresearch code for a future velocypack upgrade

jsteemann:
after the velocypack upgrade, the types of VPackXXXIterator::begin() and VPackXXXIterator::end() may differ. as end() may use `std::default_sentinel_t`.
The change in this PR should be compatible with the "old" and new versions of velocypack.

Commit that made the change in velocypack:
https://github.com/arangodb/velocypack/pull/141/files#diff-c1541b17db38faa9c3d35b5518644ce5f3ccdeeffc0a2de7ba65deb33e99bab6R90

### DIFF
--- a/core/analysis/token_stopwords_stream.cpp
+++ b/core/analysis/token_stopwords_stream.cpp
@@ -101,7 +101,8 @@ irs::analysis::analyzer::ptr construct(const VPackArrayIterator& mask,
   size_t offset = 0;
   irs::analysis::token_stopwords_stream::stopwords_set tokens;
 
-  for (auto itr = mask.begin(), end = mask.end(); itr != end; ++itr, ++offset) {
+  auto end = mask.end();
+  for (auto itr = mask.begin(); itr != end; ++itr, ++offset) {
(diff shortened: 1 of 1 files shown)


MBkkt (approved):

core/analysis/token_stopwords_stream.cpp
```diff
@@ -101,7 +101,8 @@ irs::analysis::analyzer::ptr construct(const VPackArrayIterator& mask,
   size_t offset = 0;
   irs::analysis::token_stopwords_stream::stopwords_set tokens;
 
-  for (auto itr = mask.begin(), end = mask.end(); itr != end; ++itr, ++offset) {
+  auto end = mask.end();
+  for (auto itr = mask.begin(); itr != end; ++itr, ++offset) {
```
MBkkt, line 105:
```suggestion
  for (auto itr = mask.begin(); itr.valid(); ++itr, ++offset) {
```

MBkkt:
https://github.com/iresearch-toolkit/iresearch/pull/470

PR #471, opened 2023-01-15
Few asan fixes and small refactoring

MBkkt:
Please check commit descriptions


core/formats/sparse_bitmap.cpp
```diff
@@ -379,7 +379,13 @@ struct container_iterator<BT_DENSE, true> {
         self->in_->seek(pos);
       } else {
         for (auto* prev_word = ctx.u64data - 1; !word; --word_idx) {
-          word = *--prev_word;
+          --prev_word;
+          if constexpr (AT_DIRECT_ALIGNED == Access) {
+            word = *prev_word;
```
jsteemann, line 384: I interpret `AT_DIRECT_ALIGNED` as "we can read the memory directly, and we are 100% sure that the address we read from is correctly aligned".
If this is correct, can we add an assertion, e.g. `assert(reinterpret_cast<uintptr_t>(&prev_word) % sizeof(word) == 0);`?

core/search/sort.hpp
```diff
@@ -58,85 +63,101 @@ struct score_ctx {
...
+  ~ScoreFunction() noexcept { deleter_(ctx_); }
 
-  void Reset(score_ctx* ctx, const score_f func) noexcept {
-    ctx_ = memory::to_managed<score_ctx, false>(ctx);
-    func_ = func;
+  void Reset(score_ctx& ctx, score_f score) noexcept {
+    deleter_(ctx_);
+    ctx_ = &ctx;
+    score_ = score;
+    deleter_ = &Noop;
```
jsteemann, line 130: Do you need some safeguard such as 
```
if (&ctx != ctx_) {
  deleter_(ctx_);
  ctx_ = &ctx;
  score_ = score;
  deleter_ = &Noop;
}
```
here?

core/utils/bit_packing.cpp
```diff
@@ -30,29 +30,27 @@
...
 #endif
 
-  template<int N>
-  void fastpack(const uint32_t* IRS_RESTRICT in,
-                uint32_t* IRS_RESTRICT out) noexcept {
+template<int N>
+void fastpack(const uint32_t* IRS_RESTRICT in,
+              uint32_t* IRS_RESTRICT out) noexcept {
   // 32 == sizeof(uint32_t) * 8
-  static_assert(N > 0 && N <= 32, "N <= 0 || N > 32");
```
jsteemann, line 53: :+1:

core/utils/bit_packing.cpp
```diff
@@ -30,29 +30,27 @@
...
-  // ensure all computations are constexr, i.e. no conditional jumps, no loops,
-  // no variable increment/decrement
+  static_assert(0 < N && N < 32);
+  // ensure all computations are constexpr,
+  // i.e. no conditional jumps, no loops, no variable increment/decrement
   *out |= ((*in) % (1U << N)) << (N * 0) % 32;
```
jsteemann, line 54: What's the purpose of the bitshift by `(N * 0)` (i.e. 0) bits here?

core/utils/bit_packing.cpp
```diff
@@ -252,9 +250,9 @@ template<int N>
 void fastpack(const uint64_t* IRS_RESTRICT in,
               uint64_t* IRS_RESTRICT out) noexcept {
   // 64 == sizeof(uint64_t) * 8
-  static_assert(N > 0 && N <= 64, "N <= 0 || N > 64");
-  // ensure all computations are constexr, i.e. no conditional jumps, no loops,
-  // no variable increment/decrement
+  static_assert(0 < N && N < 64);
+  // ensure all computations are constexpr,
+  // i.e. no conditional jumps, no loops, no variable increment/decrement
   *out |= ((*in) % (1ULL << N)) << (N * 0) % 64;
```
jsteemann, line 256: What's the purpose of the bitshift by `(N * 0)` (i.e. 0) bits here?

core/utils/index_utils.cpp
```diff
@@ -273,7 +273,7 @@ ConsolidationPolicy MakePolicy(const ConsolidateDocsFill& options) {
   return [options](Consolidation& candidates, const IndexReader& reader,
                    const ConsolidatingSegments& /*consolidating_segments*/) {
     auto fill_threshold = options.threshold;
-    auto threshold = std::clamp(0.0f, fill_threshold, 1.0f);
+    auto threshold = std::clamp(fill_threshold, 0.0F, 1.0F);
```
jsteemann, line 276: :+1:

core/utils/memory.hpp
```diff
@@ -255,86 +255,66 @@ class allocator_array_deleter {
...
-  friend struct managed_deleter;
+  void Destroy() const final { delete this; }
+};
 
-  pointer ptr_;
-};  // managed_deleter
+struct ManagedDeleter {
+  void operator()(const Managed* p) noexcept {
+    IRS_ASSERT(p != nullptr);  // std::unique_ptr doesn't call dtor on nullptr
+    p->Destroy();
```
jsteemann, line 298: Isn't the call to `Destroy()` here a virtual function call? Should `ManagedDeleter` be templated using the type of the specialized class it manages?

core/utils/mmap_utils.cpp
```diff
@@ -109,6 +80,11 @@ bool mmap_handle::open(const file_path_t path) noexcept {
   if (size) {
     size_ = size;
 
+    // TODO(MBkkt) Needs benchmark?
+    //  1. MAP_SHARED can makes more sense than MAP_PRIVATE
```
jsteemann, line 84:
```suggestion
    //  1. MAP_SHARED can make more sense than MAP_PRIVATE.
```

core/search/sort.hpp
MBkkt: Good catch!

core/utils/memory.hpp
MBkkt: It cannot be templated, type of deleter will be different

MBkkt: Also it always was virtual, because before my changes was call of virtual dtor

core/utils/bit_packing.cpp
MBkkt: I think it's just same way to write different iterations of this functions

core/search/sort.hpp
MBkkt: Hmm, not really
`IRS_ASSERT(&ctx != ctx_ || deleter_ == &Noop);`

core/utils/mmap_utils.cpp
MBkkt: Why?

core/search/sort.hpp
MBkkt: I will add this assert

core/search/sort.cpp
```diff
@@ -73,84 +70,59 @@ std::tuple<Vector, size_t, IndexFeatures> Prepare(Iterator begin,
...
-                                                 uint32_t size) noexcept {
-  if (0 == size) {
+ScoreFunction ScoreFunction::Constant(score_t value, uint32_t count) noexcept {
+  if (0 == count) {
     return {};
-  } else if (1 == size) {
+  } else if (1 == count) {
     return Constant(value);
   } else {
-    struct ScoreCtx {
```
gnusi, line 110: .

gnusi (approved):

core/analysis/text_token_stream.cpp
```diff
@@ -359,8 +353,8 @@ analysis::analyzer::ptr construct(icu::Locale&& locale) {
       return nullptr;
     }
 
-    return construct(options.locale.getName(), std::move(options),
-                     std::move(stopwords));
+    return construct(hashed_string_view{options.locale.getName()},
+                     std::move(options), std::move(stopwords));
```
jsteemann, line 357: Is this safe? We are creating a hased_string_view on some locale name from `options`, but the next action is to `std::move` the `options` object somewhere else.

core/utils/bit_packing.cpp
jsteemann: Maybe, but bitshifting something by 0 still seems kind of useless. Why not just remove the `<< (N * 0)` part here?

core/analysis/text_token_stream.cpp
MBkkt: Yes, it's safe because construct accept by rvalue reference, so there hasn't move ctor

MBkkt:
https://github.com/arangodb/arangodb/pull/17989

PR #483, opened 2023-03-01
Add ability to disable include execinfo

MBkkt:


core/utils/log.cpp
```diff
@@ -588,9 +598,11 @@ void stack_trace_posix(irs::logger::level_t level, int out_pipe) {
   close(pipefd[0]);
   std::fprintf(out, "\n");
   std::fflush(out);
+#endif
 }
 #endif
 
+#ifndef DISABLE_EXECINFO
```
Dronplane, line 605:
```suggestion
#if not defined(DISABLE_EXECINFO) && defined(USE_LIBBFD)
```
and remove next ifdef

Dronplane (approved):
LGTM

PR #486, opened 2023-03-18
Add ability to Remove without increment tick

MBkkt:
Before we had a requirement for the tick provider to have a separate tick for each Remove. But sometimes removes could be "fake" and does not produce a separate tick. So we added the ability to have Remove that does not require additional tick


Dronplane (approved):
LGTM

Dronplane:
DO NOT REMOVE THE BRANCH FOR NOW.

PR #491, opened 2023-03-31
Fix windows debug

MBkkt:


Dronplane (approved):
LGTM

core/search/scorer_impl.h
```diff
@@ -0,0 +1,171 @@
...
+    docs_with_term += docs_with_term_value;
+  }
+
+  void write(data_output& out) const final { out.write_vlong(docs_with_term); }
+};
+
+struct Empty final {};
+
+template<bool Freq>
+struct FieldCollectorImpl final : FieldCollector {
```
Dronplane, line 92: This class seems a half different. Maybe better to keep this implementations separated?

gnusi (approved):

PR #494, opened 2023-04-07
Add ability to reopen columnstore on Commit

MBkkt:
This feature allow us to change cache options for arangosearch view / inverted index in runtime.

And bufgix!

It's possible in old code we change existing mask file, in case of crash we will faced with inconsistency
Also we don't update version of segment


core/index/index_writer.hpp
```diff
@@ -68,21 +70,8 @@ ENABLE_BITMASK_ENUM(OpenMode);
...
-
-struct CandidateEqual {
-  bool operator()(const SubReader* lhs, const SubReader* rhs) const noexcept {
-    return lhs->Meta().name == rhs->Meta().name;
-  }
-};
-
 // segments that are under consolidation
-using ConsolidatingSegments =
-  absl::flat_hash_set<const SubReader*, CandidateHash, CandidateEqual>;
```
MBkkt, line 85: I think hash_map string_view will work better because we avoid 2 indirections

core/index/index_writer.hpp
```diff
@@ -179,6 +168,12 @@ struct IndexWriterOptions : public SegmentOptions {
   IndexWriterOptions() {}  // compiler requires non-default definition
 };
 
+struct CommitInfo {
+  uint64_t tick = writer_limits::kMaxTick;
+  ProgressReportCallback progress;
+  bool need_reopen = false;
```
MBkkt, line 174: We can reopen columnstore now

core/index/segment_reader_impl.cpp
```diff
@@ -159,124 +158,75 @@ std::vector<index_file_refs::ref_t> GetRefs(const directory& dir,
...
-    info_{meta},
-    dir_{rhs.dir_},
-    sort_{rhs.sort_},
-    opts_{rhs.opts_} {}
+void SegmentReaderImpl::Update(const directory& dir, const SegmentMeta& meta,
+                               DocumentMask&& docs_mask) noexcept {
+  IRS_ASSERT(meta.live_docs_count <= meta.docs_count);
+  IRS_ASSERT(docs_mask.size() <= meta.docs_count);
+  // TODO(MBkkt) on practice only mask file changed, so it can be optimized
+  refs_ = GetRefs(dir, meta);
```
MBkkt, line 226: I think will be nice to assume last file is mask file

core/index/index_writer.hpp
```diff
@@ -931,13 +921,14 @@ class IndexWriter : private util::noncopyable {
...
   // Start transaction
-  bool Start(uint64_t tick, const ProgressReportCallback& progress);
+  bool Start(const CommitInfo& info);
   // Finish transaction
   void Finish();
   // Abort transaction
   void Abort() noexcept;
 
   feature_set_t wand_features_;  // Set of features required for wand
+  ScorersView wand_scorers_;
```
MBkkt, line 942: Now we don't need to fetch reader to get segment writer options

core/index/index_writer.hpp
```diff
@@ -179,6 +168,12 @@ struct IndexWriterOptions : public SegmentOptions {
   IndexWriterOptions() {}  // compiler requires non-default definition
 };
 
+struct CommitInfo {
```
MBkkt, line 171: probably not best name

core/index/index_writer.hpp
MBkkt, line 174: probably not best name

core/index/index_writer.cpp
```diff
@@ -2238,20 +2217,29 @@ IndexWriter::PendingContext IndexWriter::PrepareFlush(
...
+      }
+      if (need_update) {
         // Write updated segment metadata
         index_utils::FlushIndexSegment(dir, new_segment);
-        // Refresh reader, TODO(MBkkt) move first two args?
-        segment_ctx.reader = std::make_shared<const SegmentReaderImpl>(
-          *segment_ctx.reader, new_segment.meta, std::move(document_mask));
+        // Refresh reader
+        segment_ctx.reader->Update(dir, new_segment.meta,
+                                   std::move(document_mask));
```
MBkkt: If segment was split, we need to move version forward and write new segment metadata

core/index/index_writer.cpp
```diff
@@ -2187,15 +2167,14 @@ IndexWriter::PendingContext IndexWriter::PrepareFlush(
...
           // We don't support case when segment is committed partially more than
           // one time. Because it's useless and ineffective.
           IRS_ASSERT(flushed_last_tick <= tick);
-          // find existing reader
-          reader = std::make_shared<const SegmentReaderImpl>(
-            *it->second, flushed.meta, DocumentMask{});
+          // reuse existing reader with initial meta and docs_mask
+          reader = it->second->Reopen(dir, flushed.meta, DocumentMask{});
         } else {
           reader = SegmentReaderImpl::Open(dir, flushed.meta, reader_options);
```
MBkkt, line 2177: Probably in both call we don't need GetRefs, we can make it later

core/index/index_writer.hpp
Dronplane, line 174:
```suggestion
  bool need_reopen_columnstore = false;
```

PR #496, opened 2023-04-14
I fixed this empty commit ****

MBkkt:
Few fixes:
1. If was empty commit but we need reopen_columnstore we need to recreate index reader
2. If some segment was split by commit and it was in reader in first commit, but it will be empty in second commit. It's not empty commit at all
3. If was rollback to some segment and wasn't commit in this segment writer we can reset it instead of masking

### DIFF
--- a/core/index/index_writer.cpp
+++ b/core/index/index_writer.cpp
@@ -1068,9 +1068,12 @@ void IndexWriter::SegmentContext::Rollback() noexcept {
     }
   }
 
-  // TODO(MBkkt) if committed_buffered_docs_ == 0 we can just reset writer?
   // rollback inserts located inside the writer
   auto& writer = *writer_;
+  if (committed_buffered_docs_ == 0) {
+    writer.reset();
+    return;
+  }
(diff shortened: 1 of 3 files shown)


core/index/index_writer.cpp
```diff
@@ -2271,6 +2270,15 @@ IndexWriter::PendingContext IndexWriter::PrepareFlush(const CommitInfo& info) {
 
   // only flush a new index version upon a new index or a metadata change
   if (!modified) {
+    IRS_ASSERT(readers.size() == committed_reader_size);
```
Dronplane, line 2273: Do we have test that covers this change?

MBkkt: Assert is just my paranoia.
If you about fix, yes I added test.
If you about reopen_columnstore test in arangodb for now

Dronplane: I was about test for "reopen". 
Please add tests to iresearch.

MBkkt: Done

Dronplane (approved):
LGTM

PR #500, opened 2023-04-21
Small improvements

MBkkt:
1. Write norm - freq instead of just norm, to make it smaller on disk
2. Avoid two unnecessary copies for conjunction


gnusi (approved):

Dronplane (approved):
LGTM

PR #501, opened 2023-04-22
Small improvements and change score_threshold interface

MBkkt:
~~TODO write whole skip-list max score on disk
TODO skiplist smaller than block probably should have score_threshold with max scores equal to very big float or another marker otherwise we couldn't use them~~


core/analysis/token_attributes.hpp
```diff
@@ -162,15 +162,28 @@ class attribute_provider_change final : public attribute {
...
-  std::span<const score_t> skip_scores;
+  score_t min;
+  // For disjunction/conjunction it's just sum of sub-iterators max score
+  // For iterator without score it depends on count of documents in iterator
+  // For wanderator it's max score for whole skip-list
+  // TODO(gnusi) write it to the begin of skip-list
+  // TODO(MBkkt) tail_max better here and not affect correctness
+  //  but to support it we need to know max value in the tail blocks.
+  //  Open question: how do it without read next blocks?
+  // score_t list_max;
```
MBkkt, line 181: without it impossible to choose any pivot (except == len) and set some min threshold to sub iterators

core/analysis/token_attributes.hpp
```diff
@@ -162,15 +162,26 @@ class attribute_provider_change final : public attribute {
   mutable callback_f callback_{&noop};
 };
 
-// Score threshold can be set by document consumers
+// min score set by document consumers
+// max score set by document producers
```
Dronplane, line 166:
```suggestion
// list_max/leaf_max score set by document producers

```

core/formats/wand_writer.hpp
```diff
@@ -72,9 +72,7 @@ class WandWriterImpl final : public WandWriter {
   }
 
   void WriteRoot(index_output& out) final {
-    auto max = std::max_element(score_levels_.begin(), score_levels_.end());
-    IRS_ASSERT(max != score_levels_.end());
-    ValueProducer::Write(max->value, out);
+    ValueProducer::Write(score_levels_.back().value, out);
```
Dronplane, line 75: Looks like in order to make WriteRoot work properly one needs to call SizeRoot.
Could we make WriteRoot correctness  not dependable on SizeRoot call?

core/formats/wand_writer.hpp
```diff
@@ -54,40 +56,48 @@ class WandWriterImpl final : public WandWriter {
...
   byte_type Size(size_t level) const noexcept final {
-    IRS_ASSERT(level < score_levels_.size());
+    IRS_ASSERT(level + 1 < score_levels_.size());
     const auto& entry = score_levels_[level];
     return ValueProducer::Size(entry.value);
   }
 
+  byte_type SizeRoot() noexcept final {
+    auto max = std::max_element(score_levels_.begin(), score_levels_.end());
+    IRS_ASSERT(max != score_levels_.end());
```
Dronplane, line 86: std::max_element always returns valid iterator unless range.
```suggestion
    IRS_ASSERT(!score_levels_.empty());
```

core/formats/formats_10.cpp
```diff
@@ -1975,9 +2047,37 @@ class doc_iterator : public doc_iterator_base<IteratorTraits, FieldTraits> {
...
+                   byte_type wand_index) {
+    prepare(meta, doc_in, pos_in, pay_in, wand_index);
+    if (meta.docs_count > FieldTraits::block_size()) {
+      return;
+    }
+    auto& threshold = std::get<score_threshold>(attrs_);
+    auto ctx = scorer.prepare_wand_source();
+    auto func = factory(*ctx, scorer);
+    const bool is_full_block = meta.docs_count == FieldTraits::block_size();
+    const auto was = this->doc_in_->file_pointer();
```
gnusi, line 2062: virtual call is needed only in case of full block

MBkkt: Fixed

core/formats/wand_writer.hpp
MBkkt: Fixed

core/formats/formats_10.cpp
```diff
@@ -702,6 +704,15 @@ void postings_writer_base::EndTerm(version10::term_meta& meta) {
     return;  // no documents to write
   }
 
+  const bool has_skip_list = skip_.Skip0() < meta.docs_count;
+  auto write_max_score = [&] {
+    ApplyWriters([&](auto& writer) {
```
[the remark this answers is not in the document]
MBkkt: Maybe in a separate PR with rename

Dronplane (approved):
LGTM

PR #504, opened 2023-05-03
Fix BM25 with k == 0 and small improvements

MBkkt:


core/search/bm25.hpp
```diff
@@ -69,7 +69,12 @@ class BM25 final : public irs::ScorerBase<BM25, BM25Stats> {
   static void init();  // for trigering registration in a static build
 
   BM25(float_t k = K(), float_t b = B(), bool boost_as_score = BOOST_AS_SCORE())
-  noexcept : k_{k}, b_{b}, boost_as_score_{boost_as_score} {}
+    : k_{k}, b_{b}, boost_as_score_{boost_as_score} {
+    if (k < 0.f || b < 0.f || 1.f < b) {
+      IRS_LOG_WARN(
```
Dronplane, line 74: Please remove this warning or reduce it to TRACE. It will pollute logs after update of existing instances with scorers violating this requirement.

MBkkt: It's intentionally, because user shouldn't use invalid score

Dronplane (commented):
For BM25 if  b==0 and k == 0 both `IsBM15` and `IsBM1`  will return true. With current usage this is not a problem.  But it is better to have `IsBM15` to check also k !=0

core/formats/formats.hpp
```diff
@@ -142,14 +143,26 @@ void postings_writer::releaser::operator()(term_meta* meta) const noexcept {
...
 struct field_writer {
   using ptr = std::unique_ptr<field_writer>;
 
   virtual ~field_writer() = default;
   virtual void prepare(const flush_state& state) = 0;
-  virtual void write(std::string_view name, IndexFeatures index_features,
-                     const std::map<type_info::type_id, field_id>& features,
-                     term_iterator& data) = 0;
+  virtual void write(const basic_term_reader& reader,
+                     const irs::feature_map_t& features) = 0;
```
gnusi, line 165: Looks a bit misleading as `field_meta` contains `feature_map_t`

core/store/memory_directory.cpp
```diff
@@ -133,7 +132,7 @@ int64_t memory_index_input::checksum(size_t offset) const {
     auto& buf = file_->get_buffer(last_buffer_idx);
     to_process = std::min(offset, file_->length() - buf.offset);
     crc.process_bytes(buf.data, to_process);
-    offset -= to_process;
+    // offset -= to_process; offset unused after this, so comment out it
```
gnusi, line 135: Why not just  remove the line then?

core/formats/formats.hpp
MBkkt: We need to pass different feature map in the consolidation

core/store/memory_directory.cpp
MBkkt: ¯⁠\⁠_⁠(⁠ツ⁠)⁠_⁠/⁠¯

gnusi (approved):

gnusi (approved):

Dronplane (approved):
LGTM

PR #508, opened 2023-05-16
Bug fix 1.3/fix wand data skip

Dronplane:
Backport https://github.com/iresearch-toolkit/iresearch/pull/507
               https://github.com/iresearch-toolkit/iresearch/pull/504


MBkkt (approved):
Jenkins run is unnecessary because master is same

PR #511, opened 2023-05-22
Term iterators improvements

MBkkt:


core/formats/formats_10.cpp
```diff
@@ -2842,7 +2843,7 @@ doc_id_t wanderator<IteratorTraits, FieldTraits, WandIndex, WandExtent,
       std::get<position<IteratorTraits, FieldTraits>>(attrs_).notify(notify);
     }
 
-    target = doc.value + 1;
+    seek_target = doc_value + 1;
```
MBkkt, line 2846: We will only decrease the current target in the "tail" case, but in this case it doesn't matter.

core/formats/formats_10.cpp
```diff
@@ -2813,27 +2812,26 @@ doc_id_t wanderator<IteratorTraits, FieldTraits, WandIndex, WandExtent,
...
-        // FIXME(gnusi): can we use approximation before actual score
-        //               evaluation? e.g. frequency in case of tfidf
-        scorer_(&score_);
-
-        if (ScoreLess(score_, min_competitive_score)) {
-          continue;
-        }
+          if (ScoreLess(score_, min_competitive_score)) {
+            continue;
+          }
```
MBkkt, line 2829: We don't need to compute score if it's unnecessary (same as without position)

core/formats/formats_10.cpp
```diff
@@ -2758,19 +2761,19 @@ template<typename IteratorTraits, typename FieldTraits, typename WandIndex,
          typename WandExtent, bool Strict>
 doc_id_t wanderator<IteratorTraits, FieldTraits, WandIndex, WandExtent,
                     Strict>::seek(doc_id_t target) {
-  auto& doc = std::get<document>(attrs_);
+  auto& doc_value = std::get<document>(attrs_).value;
```
MBkkt, line 2764: we don't access doc, so instead of twice indirection make it single time

core/formats/formats_10.cpp
MBkkt: It's probably fix seek in the tail for this iterator

MBkkt: I don't know how, in general it's in the same case when while needed for next in the doc_iterator

Dronplane (approved):
LGTM

PR #512, opened 2023-05-23
Remove not strict wand

MBkkt:


tests/index/assert_format.cpp
```diff
@@ -746,7 +746,7 @@ void assert_docs(const irs::term_iterator& expected_term,
       }};
 
     return actual_terms.wanderator(*actual_cookie, requested_features, options,
-                                   {.index = 0, .strict = false});
+                                   {.index = 0});
```
MBkkt, line 749: Looks like it doesn't matter in this case

MBkkt, line 749: It looks like no difference for this test

tests/search/wand_test.cpp
```diff
@@ -119,6 +119,7 @@ std::vector<Doc> WandTestCase::Collect(const irs::DirectoryReader& index,
                                        const irs::Scorer& scorer,
                                        irs::byte_type wand_idx,
                                        bool can_use_wand, size_t limit) {
+  constexpr bool strict = false;
```
MBkkt, line 122: It was false in the line 127

PR #515, opened 2023-05-26
Fix typo

MBkkt:


Dronplane (changes requested):
Please add a test to cover this case. As it looks like current tests seems to ignore this case.

Dronplane (approved):
LGTM

PR #516, opened 2023-05-26
Try less memory usage

MBkkt:
0.4 vs 0.5 on my machine with old memset.
And 0.1 without unnecessary memset

### DIFF
--- a/core/formats/columnstore2.hpp
+++ b/core/formats/columnstore2.hpp
@@ -89,44 +89,33 @@ class column final : public irs::column_output {
   class address_table {
    public:
     uint64_t back() const noexcept {
-      IRS_ASSERT(offset_ > offsets_);
-      return *(offset_ - 1);
+      IRS_ASSERT(!empty());
+      return offsets_.back();
     }
 
-    void push_back(uint64_t offset) noexcept {
-      IRS_ASSERT(offset_ >= offsets_);
-      IRS_ASSERT(offset_ < offsets_ + kBlockSize);
-      *offset_++ = offset;
-      IRS_ASSERT(offset >= offset_[-1]);
-    }
+    void push_back(uint64_t offset) { offsets_.push_back(offset); }
 
     void pop_back() noexcept {
-      IRS_ASSERT(offset_ > offsets_);
-      *--offset_ = 0;
+      IRS_ASSERT(!empty());
+      offsets_.pop_back();
     }
 
-    // returns number of items to be flushed
-    uint32_t size() const noexcept {
-      IRS_ASSERT(offset_ >= offsets_);
-      return uint32_t(offset_ - offsets_);
+    uint32_t size() const noexcept { return offsets_.size(); }
+
+    bool empty() const noexcept { return offsets_.empty(); }
+
+    bool full() const noexcept { return offsets_.size() == kBlockSize; }
+
+    void reset() noexcept { offsets_.clear(); }
+
+    uint64_t* begin() noexcept { return offsets_.data(); }
+    uint64_t* current() noexcept {
+      IRS_ASSERT(!empty());
+      return begin() + offsets_.size() - 1;
     }
-
-    bool empty() const noexcept { return offset_ == offsets_; }
-
-    bool full() const noexcept { return offset_ == std::end(offsets_); }
-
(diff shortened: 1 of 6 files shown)


core/formats/columnstore2.hpp
```diff
@@ -82,45 +82,46 @@ class column final : public irs::column_output {
 
   class address_table {
    public:
+    address_table(ManagedTypedAllocator<uint64_t> alloc) : alloc_{alloc} {
+      offsets_ = alloc_.allocate(kBlockSize);
```
MBkkt, line 86: Now we allocate very good aligned blocks, so I think it's better utilize allocator.

core/formats/columnstore2.hpp
```diff
@@ -82,45 +82,46 @@ class column final : public irs::column_output {
...
       IRS_ASSERT(offset_ < offsets_ + kBlockSize);
       *offset_++ = offset;
-      IRS_ASSERT(offset >= offset_[-1]);
     }
 
     void pop_back() noexcept {
-      IRS_ASSERT(offset_ > offsets_);
-      *--offset_ = 0;
+      IRS_ASSERT(offsets_ < offset_);
+      --offset_;
```
MBkkt, line 104: we don't need to set zero to unused data

core/formats/columnstore2.hpp
```diff
@@ -82,45 +82,46 @@ class column final : public irs::column_output {
...
+      return static_cast<uint32_t>(offset_ - offsets_);
     }
 
     bool empty() const noexcept { return offset_ == offsets_; }
 
-    bool full() const noexcept { return offset_ == std::end(offsets_); }
+    bool full() const noexcept { return offset_ == offsets_ + kBlockSize; }
 
-    void reset() noexcept {
-      std::memset(offsets_, 0, sizeof offsets_);
```
MBkkt, line 113: we don't need to set zero to unused data

core/formats/columnstore2.hpp
```diff
@@ -82,45 +82,46 @@ class column final : public irs::column_output {
...
+    void reset() noexcept { offset_ = offsets_; }
 
-    uint64_t* begin() noexcept { return std::begin(offsets_); }
+    uint64_t* begin() noexcept { return offsets_; }
     uint64_t* current() noexcept { return offset_; }
-    uint64_t* end() noexcept { return std::end(offsets_); }
+    uint64_t* end() noexcept { return offsets_ + kBlockSize; }
 
    private:
-    uint64_t offsets_[kBlockSize]{};
```
MBkkt, line 122: We don't need zero unused data

core/formats/columnstore2.hpp
```diff
@@ -82,45 +82,46 @@ class column final : public irs::column_output {
 
   class address_table {
    public:
+    address_table(ManagedTypedAllocator<uint64_t> alloc) : alloc_{alloc} {
+      offsets_ = alloc_.allocate(kBlockSize);
+      offset_ = offsets_;
+    }
+
+    ~address_table() { alloc_.deallocate(offsets_, kBlockSize); }
```
MBkkt, line 90: Also deque with columns works better now:
blocks in deque could handle few columns and we don't store this memory forever.
only while write segment

core/formats/columnstore2.hpp
Dronplane: Here would be 2 allocations. One for deque itself and this one for internal contents. Better to benchmark this on test load with many columns.

microbench/CMakeLists.txt
```diff
@@ -76,3 +76,21 @@ target_include_directories(iresearch-memory-file-bench
 target_link_libraries(iresearch-memory-file-bench
   iresearch-static
   )
+
+add_executable(iresearch-column-writer
+  column_writer_benchmark.cpp
```
Dronplane, line 81: Missing file?

microbench/column_writer_benchmark.cpp
```diff
@@ -0,0 +1,35 @@
+#include <random>
```
Dronplane, line 1: Disclaimer missing

Dronplane (approved):
LGTM

PR #518, opened 2023-05-31
shit

MBkkt:


core/store/mmap_directory.cpp
```diff
@@ -104,6 +104,40 @@ std::shared_ptr<mmap_handle> OpenHandle(const std::filesystem::path& dir,
   return nullptr;
 }
 
+#ifdef __linux__
+IRS_NO_INLINE size_t BytesInCache(uint8_t* addr, size_t length) {
+  static const size_t kPageSize = sysconf(_SC_PAGESIZE);
+  IRS_ASSERT(reinterpret_cast<uintptr_t>(addr) % kPageSize == 0);
+  std::vector<uint8_t> pages(
```
gnusi, line 111: I suggest to use `std::array` instead

core/store/data_input.hpp
```diff
@@ -109,6 +109,12 @@ struct data_input {
   data_input& operator++(int) noexcept { return *this; }
 };
 
+struct MemoryStats {
+  uint64_t* fd_count{};
+  uint64_t* mmaped_memory{};
+  uint64_t* pinned_memory{};
```
gnusi, line 115: TBH I don't see any value in having stats as pointers.

core/store/data_input.hpp
```diff
@@ -142,6 +148,8 @@ struct index_input : public data_input {
   //////////////////////////////////////////////////////////////////////////////
   virtual int64_t checksum(size_t offset) const = 0;
 
+  virtual void CountMemory(MemoryStats stats) const {}
```
gnusi, line 151: We also have to implement it for fs/memory

core/store/mmap_directory.cpp
MBkkt: vector is better for small mapping, especially if we will make bigger max size

core/store/data_input.hpp
MBkkt: It should be pointer or reference, reference is less flexible

core/store/data_input.hpp
MBkkt: Ok

core/store/data_input.hpp
Dronplane: The plan is to "opt-out" collecting some data, if necessary, by providing nullptr.

core/store/mmap_directory.cpp
```diff
@@ -104,6 +104,40 @@ std::shared_ptr<mmap_handle> OpenHandle(const std::filesystem::path& dir,
   return nullptr;
 }
 
+#ifdef __linux__
+IRS_NO_INLINE size_t BytesInCache(uint8_t* addr, size_t length) {
```
Dronplane, line 108:
```suggestion
size_t BytesInCache(uint8_t* addr, size_t length) {
```

core/formats/columnstore.cpp
```diff
@@ -2307,6 +2307,10 @@ class reader final : public columnstore_reader, public context_provider {
  public:
   explicit reader(size_t pool_size = 16) : context_provider(pool_size) {}
 
+  void CountMemory(MemoryStats stats) const final {
+    // TODO(MBkkt) should we implement it for old columnstore?
```
Dronplane, line 2311: I'd suggest no. We are going to deprecate it sooner or later.

core/index/index_reader.hpp
```diff
@@ -129,6 +129,8 @@ struct SubReader : public IndexReader, public ColumnProvider {
 
   size_t size() const noexcept final { return 1; }
 
+  virtual void CountMemory(MemoryStats stats) const = 0;
```
Dronplane, line 132: Why not move it to "IndexReader" ?

Dronplane: And would be better to pass value by reference. Copying 3 pointers now is not a big problem. But this stats might grow in future.

MBkkt: I think now is better by value.
If size of struct will growth a lot, we could change this

MBkkt: Ok

Dronplane (approved):
LGTM

PR #519, opened 2023-06-01
Fix skiplist iteration 

Dronplane:
Fix skiplist iteration when there is possibility to go up and down by levels (WAND case)


core/formats/formats_10.cpp
```diff
@@ -2653,6 +2653,7 @@ template<typename IteratorTraits, typename FieldTraits, typename WandIndex,
 size_t wanderator<IteratorTraits, FieldTraits, WandIndex, WandExtent,
                   Strict>::ReadSkip::AdjustLevel(size_t level) const noexcept {
   while (level && skip_levels_[level].doc >= skip_levels_[level - 1].doc) {
+    IRS_ASSERT(skip_levels_[level - 1].doc != doc_limits::eof());
```
MBkkt, line 2656:
```suggestion
  IRS_ASSERT(skip_levels_[level].doc != doc_limits::eof());
  while (level && skip_levels_[level].doc >= skip_levels_[level - 1].doc) {
```

PR #521, opened 2023-06-06
[SEARCH-476] [APM-757] Fix fst builder

MBkkt:
How does it work before?

When I looked to the failed assert it starts to looks obviously that we are doing it wrong
`DivideLeft("\x04\x03G?\x02\x05S?\x04\x05i?\x06\x05", "\x05");` 

From intuitive point of view we are trying to reduce initial output.
To do this we put part of it to the outputs of the common prefix (prev_key, curr_key).
To do this we need to cut common prefix of the current weight(output) and previous output and put suffix to the arcs.
Then make previous output prefix.
Then reduce current output to the written prefix.

I also checked reference paper and it also using CurrentOutput and other implementation of this algorithm

https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.24.3698&rep=rep1&type=pdf
image
https://github.com/phiresky/tantivy-fst/blob/dynamic-loading/src/raw/build.rs#L400

### DIFF
--- a/core/utils/fstext/fst_builder.hpp
+++ b/core/utils/fstext/fst_builder.hpp
@@ -134,7 +134,7 @@ class fst_builder : util::noncopyable {
       auto& last_out = p.arcs.back().out;
 
       if (last_out != weight_t::One()) {
-        auto prefix = fst::Plus(last_out, out);
+        auto prefix = fst::Plus(last_out, output);
         const auto suffix = fst::DivideLeft(last_out, prefix);
 
         for (arc& a : s.arcs) {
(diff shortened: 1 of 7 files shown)


MBkkt:
https://jenkins.arangodb.biz/view/PR/job/iresearch-pr/892/
https://jenkins01.arangodb.biz/view/PR/job/iresearch-debug-pr/108/

MBkkt (commented):
I also really want to remove all fst abstractions and other stuff because now in this cycle instead of just resize to smaller size in the case of the DivideLeft we are copying strings :(

Dronplane (approved):
LGTM

Dronplane (approved):
LGTM

PR #524, opened 2023-06-08
wand improvements

MBkkt:
* Don't compute real score for tfidf, bm15, bm11 on index stage
* Add valid for any BM25 less effective option for wand data producer: max tf, min norm
* Don't allow to write wand data for BM1 (constant score, just idf)

Also optimize fst builder: don't need to copy itself prefix

### DIFF
--- a/core/formats/wand_writer.hpp
+++ b/core/formats/wand_writer.hpp
@@ -28,16 +28,18 @@
 #include "index/norm.hpp"
 #include "search/scorer.hpp"
 #include "store/memory_directory.hpp"
-
-#include <absl/container/inlined_vector.h>
+#include "utils/empty.hpp"
+#include "utils/small_vector.hpp"
 
 namespace irs {
 
-template<typename ValueProducer>
+template<typename Producer>
 class WandWriterImpl final : public WandWriter {
+  using EntryType = typename Producer::Entry;
+
  public:
   WandWriterImpl(const Scorer& scorer, size_t max_levels)
-    : score_levels_{max_levels + 1}, producer_{scorer} {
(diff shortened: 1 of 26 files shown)


core/formats/wand_writer.hpp
```diff
@@ -47,238 +49,295 @@ class WandWriterImpl final : public WandWriter {
...
 
-class ValueProducerBase {
- public:
-  void GetScore(score_t* score) const noexcept { func_(score); }
-
-  explicit ValueProducerBase(const Scorer& scorer) : scorer_{&scorer} {
+template<>
+struct WandScorer<true> {
+  explicit WandScorer(const Scorer& scorer) : scorer_{scorer} {
+    // TODO(MBkkt) alignment could be incorrect here
```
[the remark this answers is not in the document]
MBkkt: I don't agree at least because sso, but I don't think it's matter now

Dronplane (approved):
LGTM fix for bit_union

tests/search/wand_test.cpp
```diff
@@ -252,6 +253,28 @@ void WandTestCase::GenerateSegment(irs::ScorersView scorers, bool write_norms,
...
+
+      if (tests::json_doc_generator::ValueType::STRING == data.vt) {
+        doc.indexed.push_back(std::make_shared<TextField>(
+          std::string{name}, data.str, false,
+          std::vector{irs::type<irs::Norm2>::id()}));
+      }
+    });
+
+  auto open_mode = irs::OM_CREATE;
+  // if (append_data) {
```
Dronplane, line 271: Is this needed?

MBkkt: Nope

PR #529, opened 2023-06-23
Remove memory allocator

MBkkt:
Firstly we never use it with non zero pool size.
So it's useless now.

Secondly it looks more like possible pessimization because internal system allocator structures a lot of better than just two lock-free lists


core/utils/container_utils.hpp
```diff
@@ -27,339 +27,146 @@
...
-    }
-
-    // non-precomputed buckets, offset is the sum of previous buckets
-    IRS_ASSERT(
-      !base_t::buffers_
-         .empty());  // otherwise do not know what size buckets to create
-    const auto& bucket =
-      base_t::buffers_.back();  // most of the meta from last computed bucket
-    return base_t::push_buffer(bucket.offset + bucket.size, META.back());
+  const buffer_t& push_buffer() {
```
Dronplane, line 147: why it is not in base class? Looks a bit strange that base has "clear" but not "push"

MBkkt: Because it's need META

Dronplane: But then the question - why we need to keep `raw_block_vector_base` ? It is not used anywhere else. Could we just get rid of it and put everything in `raw_block_vector`?

MBkkt: Same as old reason: to instantiate over methods single time

Dronplane: then why not with "push_buffer" part that does not depends on META ?

Dronplane (approved):
LGTM

PR #531, opened 2023-06-27
Fixed: O(n) seek in buffered column. Benchmarks great again.

MBkkt:


tests/index/buffered_column_test.cpp
```diff
@@ -178,12 +178,12 @@ void AssertIteratorCornerCases(const irs::BufferedColumn& column,
...
-    ASSERT_FALSE(irs::doc_limits::valid(it.seek(irs::doc_limits::invalid())));
+    const auto value = it.seek(irs::doc_limits::invalid());
+    ASSERT_EQ(value, it.value());
     if (!expected_values.empty()) {
-      ASSERT_TRUE(it.next());
-      ASSERT_EQ(1, it.value());
+      ASSERT_EQ(value, 1);
     } else {
-      ASSERT_FALSE(it.next());
+      ASSERT_EQ(value, irs::doc_limits::eof());
```
Dronplane, line 186: Why dropped check for `next` after eof? This is valid and well defined 
// Once iterator is exhausted:
//   - `next()` must constantly return `false`
//   - `seek()`, `shallow_seek()` to any value must return `doc_limits::eof()`
//   - `value()` must return `doc_limits::eof()`

Dronplane (commented):
LGTM. See note about test

tests/index/buffered_column_test.cpp
MBkkt: This test already has such check

Dronplane (approved):
LGTM

PR #540, opened 2023-07-29
Optimize bitset.all()

MBkkt:
Add early return

### DIFF
--- a/core/utils/bitset.hpp
+++ b/core/utils/bitset.hpp
@@ -198,7 +198,19 @@ class dynamic_bitset {
 
   bool none() const noexcept { return !any(); }
 
-  bool all() const noexcept { return count() == bits_; }
+  bool all() const noexcept {
+    if (words_ == 0) {
+      return true;
+    }
+    auto* begin = data_.get();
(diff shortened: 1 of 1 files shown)


core/utils/bitset.hpp
```diff
@@ -198,7 +198,19 @@ class dynamic_bitset {
...
+      return true;
+    }
+    auto* begin = data_.get();
+    for (auto* end = begin + words_ - 1; begin != end; ++begin) {
+      static_assert(std::is_unsigned_v<word_t>);
+      if (*begin != std::numeric_limits<word_t>::max()) {
+        return false;
+      }
+    }
+    return std::popcount(*begin) == bits_;
```
Dronplane, line 212: Should be something like
```suggestion
    return std::popcount(*begin) == bits_ - (words_-1) * sizeof(word_t);
```

Are we missing a test for such case?

Dronplane (approved):
LGTM in general. But please add a test for counting last word, as look like this issue happens and we don't cover it with existing tests.

MBkkt:
We already have such tests. It's all cases when bitset contains only 1 word

PR #541, opened 2023-07-31
Reduce burst trie memory usage

MBkkt:
Main improvements are
* clear monotonic buffer after field end
* release monotonic buffer after writes end

but always keep next block size, also

* sizeof prefixed_output 96 instead of 104 (node in list)
  because single linked list instead of double linked, so it single pointer in the Node instead of two
* sizeof entry(block_index_t) 88(8) instead of 120(40)
  because list just a single tail pointer instead of head, tail, size and alloc

### DIFF
--- a/core/formats/formats_burst_trie.cpp
+++ b/core/formats/formats_burst_trie.cpp
@@ -22,22 +22,11 @@
 
 #include "formats_burst_trie.hpp"
 
-#include <list>
 #include <variant>
 
 #include "utils/assert.hpp"
 
 // clang-format off
-
-#if (defined(__clang__) || defined(_MSC_VER) || \
-     defined(__GNUC__) &&                       \
-       (__GNUC__ > 8))  // GCCs <=  don't have "<version>" header
-#include <version>
-#endif
-
-#ifdef __cpp_lib_memory_resource
-#include <memory_resource>
-#endif
 
 #include "utils/fstext/fst_utils.hpp"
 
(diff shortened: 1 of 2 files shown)


core/formats/formats_burst_trie.cpp
```diff
@@ -175,11 +164,81 @@ using volatile_byte_ref = volatile_ref<byte_type>;
...
+      tail_ = &front;
+      return;
+    }
+    front.next = tail_->next;
+    tail_->next = &front;
+  }
+
+  template<typename Func>
+  IRS_FORCE_INLINE void Visit(Func&& func) const {
+    if (IRS_LIKELY(tail_ == nullptr)) {
```
Dronplane, line 222: Does this really "likely" that we visit empty lists?

core/formats/formats_burst_trie.cpp
```diff
@@ -175,11 +164,81 @@ using volatile_byte_ref = volatile_ref<byte_type>;
...
+    // h1->t1->h2->t2->h1
+    rhs.tail_->next = head;
+    // h1->**->h2->t2/t1->h1
+    tail_ = rhs.tail_;
+    // h1->**->h2->t1->h1
+    rhs.tail_ = nullptr;
+  }
+
+  void PushFront(T& front) noexcept {
+    IRS_ASSERT(front.next == &front);
```
Dronplane, line 211: Why we need this requirement? Isn't it safer to require "nullptr" in `next` for any incoming node to ensure it does not belongs to any other list. And then here set the right `next` value. BTW this assert will not fire if we have other list with single node and will pass by error that node here.

MBkkt: I know. Because it's classic way to write circular list

core/formats/formats_burst_trie.cpp
MBkkt: For us yes

core/formats/formats_burst_trie.cpp
```diff
@@ -225,16 +274,116 @@ struct block_t : private util::noncopyable {
...
+    // TODO(MBkkt) use allocate_at_least but it's only C++23 :(
+    resource_manager_.Increase(size);
+    blocks_memory_ += size;
+    auto* p =
+      static_cast<byte_type*>(operator new(size, std::align_val_t{kAlign}));
+    head_ = new (p) Block{head_};
+    IRS_ASSERT(p == reinterpret_cast<byte_type*>(head_));
+    current_ = p + sizeof(Block);
+    available_ = next_size_;
+    next_size_ = (next_size_ * 3) / 2;
```
Dronplane, line 373: We could track first call of `AllocateMemory`   and do increase only if not first. So we will not need tricks with re-counting the size.

MBkkt: I don't like if approach because it more cases to think

PR #543, opened 2023-08-10
Fix some dangerous code

MBkkt:


core/index/index_writer.hpp
```diff
@@ -733,22 +733,31 @@ class IndexWriter : private util::noncopyable {
 
  private:
   struct SegmentLimits {
+   private:
+    // TODO(MBkkt) Change zero meaning
+    static constexpr auto kSizeMax = std::numeric_limits<size_t>::max();
+    static constexpr auto kDocsMax = std::numeric_limits<uint32_t>::max() - 2;
+    static auto ZeroMax(auto value, auto max) noexcept {
```
Dronplane, line 740: We can, at least, make `max`  as `real max` -1  to save 1 subtraction as this is anyway deep private stuff

Dronplane (approved):
LGTM.

core/index/index_writer.hpp
MBkkt: It's compile time substraction

PR #544, opened 2023-08-11
Remove some unused code

MBkkt:


tests/utils/memory_tests.cpp
```diff
@@ -32,6 +32,142 @@
...
+namespace irs::memory {
+
+///////////////////////////////////////////////////////////////////////////////
+/// @class memory_multi_size_pool
+/// @brief an allocator that can handle allocations of slots with different
+///        sizes
+///////////////////////////////////////////////////////////////////////////////
+template<typename GrowPolicy = log2_grow,
+         typename BlockAllocator = malloc_free_allocator>
+class memory_multi_size_pool : public pool_base<GrowPolicy, BlockAllocator> {
```
Dronplane, line 44: If this is now test only code - why not remove/modify tests that use it?

MBkkt: I think it can be done separately.
Also this test code could not directly but test base classes, which is still used

MBkkt: About memory pool I think it can be removed at all, I just needs benchmark

Dronplane (approved):
LGTM

PR #551, opened 2023-08-20
Fix warnings

MBkkt:
```
add_compile_options(
  -Weverything
  -Wno-c++98-compat
  -Wno-c++98-compat-pedantic
  -Wno-c++20-compat
  -Wno-documentation
  -Wno-float-equal
  -Wno-sign-conversion
  -Wno-reserved-macro-identifier
  -Wno-weak-vtables
  -Wno-shadow-field-in-constructor
  -Wno-exit-time-destructors
  -Wno-padded
  -Wno-global-constructors
  -Wno-implicit-int-conversion
  -Wno-ctad-maybe-unsupported
  -Wno-used-but-marked-unused
  -Wno-defaulted-function-deleted
  -Wno-deprecated-copy-with-dtor
  -Wno-shadow-uncaptured-local
  -Wno-nested-anon-types
  -Wno-shadow
  -Wno-shadow-field
  -Wno-missing-field-initializers
  -Wno-gnu-anonymous-struct
  -Wno-date-time
  -Wno-disabled-macro-expansion
  -Wno-switch-enum
  -Wno-frame-larger-than
  -Wno-documentation-unknown-command
  -Wno-old-style-cast # icu
  -Wno-missing-prototypes # only benchmarks
)
```


core/utils/file_utils.cpp
```diff
@@ -872,8 +854,8 @@ handle_t open(void* file, OpenMode mode, int advice) noexcept {
   // under /proc/self/fd the link is guaranteed to point to the original inode
   // even if the original file was removed
   auto fd = handle_cast(file);
-  char path[strlen("/proc/self/fd/") + sizeof(fd) * 3 +
-            1];  // approximate maximum number of chars, +1 for \0
+  // approximate maximum number of chars, +1 for \0
```
Dronplane, line 857: NIT
we can make `/proc/self/fd/%d` a `constexpr string_view` and use `size()-2` instead of magic number "14"and `data()` below

Dronplane (approved):
LGTM

core/utils/file_utils.cpp
MBkkt: Will be done in the next PR

PR #554, opened 2023-09-06
Fix build with liburing

MBkkt:


core/store/async_directory.cpp
```diff
@@ -466,10 +466,12 @@ void AsyncIndexOutput::write_bytes(const byte_type* b, size_t length) {
...
-AsyncDirectory::AsyncDirectory(std::filesystem::path path,
-                               directory_attributes attrs, size_t pool_size,
-                               size_t queue_size, unsigned flags)
-  : MMapDirectory{std::move(path), std::move(attrs)},
+AsyncDirectory::AsyncDirectory(std::filesystem::path dir,
+                               directory_attributes attrs,
+                               const ResourceManagementOptions& /*rm*/,
+                               size_t pool_size, size_t queue_size,
+                               unsigned flags)
+  : MMapDirectory{std::move(dir), std::move(attrs)},
```
Dronplane, line 474: Why not pass resource manager to the underlying mmap directory?

Dronplane: That is just one-line change. Please fix ) Otherwise we will just forget about it.

Dronplane (approved):
LGTM

PR #558, opened 2023-09-12
Avoid char aliasing

MBkkt:


MBkkt:
I don't see significant performance difference

PR #561, opened 2023-09-16
Speedup segmentation

MBkkt:
3 ideas:
1. We don't need to call `std::next(it) multiple time
2. We don't need next_next
3. We can simplify check if we rearrange enum


Dronplane (approved):
LGTM

PR #562, opened 2023-09-20
Fixed default cmake behaviour

kvahed:
On Ubuntu, Debian, RedHat default behaviour of `cmake ..` is now success. Provided that the libraries are installed.


external/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 include(UpdateModule)
+include(CheckCCompilerFlag)
```
MBkkt, line 2: Why does it needed?

MBkkt (approved):

PR #569, opened 2023-10-23
attempt to fix compile warning

jsteemann:
the compile option `-fsized-deallocation` should only be used for C++ code, but not for C code. otherwise clang warns with the following message:
```
warning: command-line option ‘-fsized-deallocation’ is valid for C++/ObjC++ but not for C
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,8 +145,8 @@ if (MSVC)
     set(CMAKE_CXX_FLAGS "/Zc:static_assert- ${CMAKE_CXX_FLAGS}")
   endif ()
 else ()
-  # We want to force clang to use it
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -145,8 +145,8 @@ if (MSVC)
     set(CMAKE_CXX_FLAGS "/Zc:static_assert- ${CMAKE_CXX_FLAGS}")
   endif ()
 else ()
-  # We want to force clang to use it
-  add_compile_options(-fsized-deallocation)
+  # We want to force clang/gcc to use it, but only for C++ files
```
MBkkt, line 148: gcc doesn't need to be forced, it's default behavior in such case

MBkkt (approved):

PR #572, opened 2023-11-02
Separate WaitGroup

MBkkt:


core/utils/wait_group.hpp
```diff
@@ -0,0 +1,76 @@
...
+// TODO(MBkkt) Considered to replace with YACLib
+struct WaitGroup {
+  explicit WaitGroup(size_t counter = 0) noexcept : counter_{counter + 1} {}
+
+  void Add(size_t counter = 1) noexcept {
+    counter_.fetch_add(counter, std::memory_order_relaxed);
+  }
+
+  void Done(size_t counter = 1) noexcept {
+    if (counter_.fetch_sub(counter, std::memory_order_acq_rel) == 1) {
```
Dronplane, line 40:
```suggestion
    if (counter_.fetch_sub(counter, std::memory_order_acq_rel) == counter) {
```
?

Dronplane (changes requested):
Why not use "counter*2" as delta values for `Add` and `Done`? And keep "1" as wait indicator. That will give us always accurate `Count`.

MBkkt:
> Why not use "counter*2" as delta values for Add and Done? And keep "1" as wait indicator. That will give us always accurate Count.

We don't need "accurate" count.

core/utils/wait_group.hpp
MBkkt: It's incorrect

Dronplane:
> > Why not use "counter*2" as delta values for Add and Done? And keep "1" as wait indicator. That will give us always accurate Count.
> 
> We don't need "accurate" count.

Why we need method that intentionally lies?

MBkkt:
> Why we need method that intentionally lies?

it's not lies

Dronplane:
https://172.26.1.0/view/PR/job/iresearch-debug-pr/211/
https://172.26.1.0/view/PR/job/iresearch-pr/997/

Dronplane (approved):
LGTM

PR #576, opened 2023-11-21
Feature/backport pool

Dronplane:
Backport new thread pool implementation

### DIFF
--- a/core/utils/async_utils.cpp
+++ b/core/utils/async_utils.cpp
@@ -67,7 +67,7 @@ void ThreadPool<UseDelay>::start(size_t threads,
   IRS_ASSERT(threads);
   threads_.reserve(threads);
   for (size_t i = 0; i != threads; ++i) {
-    threads_.emplace_back([&] {
+    threads_.emplace_back([this, name] {
       if (!name.empty()) {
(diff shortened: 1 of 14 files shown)


core/utils/async_utils.cpp
```diff
@@ -55,305 +54,102 @@ void busywait_mutex::unlock() noexcept {
...
+template<bool UseDelay>
+void ThreadPool<UseDelay>::start(size_t threads,
+                                 std::basic_string_view<Char> name) {
+  IRS_ASSERT(threads_.empty());
+  IRS_ASSERT(threads);
+  threads_.reserve(threads);
+  for (size_t i = 0; i != threads; ++i) {
+    threads_.emplace_back([&] {
+      if (!name.empty()) {
+        set_thread_name(name.data());
```
MBkkt, line 72: It's not really valid, but it doesn't really matter for us, because we use only names that known in compile time.
Threads can be scheduled, but not runned, start function ended, name isn't valid, and only than threads run

MBkkt (approved):

Dronplane:
https://jenkins.arangodb.biz/view/PR/job/iresearch-pr/1003/

PR #579, opened 2023-11-30
Wildcard preparation 2

MBkkt:
* Allow to create boolean queries without boolean filters
* Allow to create ngram query without ngram options
* Optimize ngram query in case when min_match_count is max_match_count


Dronplane (approved):
LGTM

PR #582, opened 2023-12-12
Fixes for utf-8

MBkkt:


core/analysis/text_token_stream.cpp
```diff
@@ -1053,18 +1053,15 @@ bool text_token_stream::next_word() {
...
       while (begin < end) {
         const auto cp = data->char32At(begin);
-
         // icu::UnicodeString::kInvalidUChar is private
-        if (IRS_UNLIKELY(0xFFFF == cp)) {
-          return uint32_t{0};
-        }
-
-        length += utf8_utils::cp_length(cp);
+        IRS_ASSERT(cp != 0xFFFF);
```
Dronplane, line 1063: Can we be sure that `icu::UnicodeString::fromUTF8` always returns valid string? From comment above it looks like it might set icu::UnicodeString::kInvalidUChar in the string in case of conversion error.

MBkkt: Fixed

Dronplane (approved):
LGTM

MBkkt:
Flaky windows tests

